PR #111, opened 2021-08-12
Premium create unlimited aliases

maxxcrawford:
This PR detects if a authenticated user is premium. If so, the max alias count restriction is lifted! 

image

## Testing

Premium user: 
- Run add-on, log into site as a premium user
- Have more than 5 aliases created on your account
- **Expected:** On email login, badge should allow you to create additional aliases. The text should read "You have created `X` aliases"

Free user: 
- Run add-on, log into site as a premium user
- Have more than 5 aliases created on your account
- **Expected:** On email login, badge should allow you to create additional aliases. The text should read "You have created `X` aliases"

## Edge cases: 

If for some reason the user sets their add-on to "premium", when they aren't – they will get the following message: 

image

If a user has created more than the max aliases and downgraded their account, the following happens: 
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fx-private-relay-add-on",
-  "version": "1.5.1",
+  "version": "1.6.0",
   "description": "Firefox Relay",
   "main": "index.js",
   "dependencies": {
--- a/src/_locales/en/messages.json
+++ b/src/_locales/en/messages.json
@@ -60,6 +60,15 @@
             }
         }
     },
+    "popupUnlimitedAliases": {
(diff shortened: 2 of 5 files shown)


codemist (approved):

groovecoder (approved):
Spot-checks work. This is a good first fix. The length of the `addRelayIconToInput` function smells like it could use some refactoring and eventually maybe considering full-on UI framework like React.

The `is-premium` check feels a bit sloppy, but as you discovered, the back-end will enforce "real" premium subscription on the account so even if a savvy add-on user tries to hack their add-on to give them premium, it won't actually work.

PR #114, opened 2021-08-23
Add Premium Panel State to Add On

codemist:
This PR implements the new premium add on design:

- Adds dashboard data from profile page into add on
- Add CTA to get custom domain from add on
- When user is premium, switch completely to a new panel that displays premium content

WIP:
- Switching from the get custom domain CTA + illustration module to the educational module if a user has a custom domain
- Need to merge with https://github.com/mozilla/fx-private-relay-add-on/pull/113 to eliminate the panel status bar that's still visible at the top when user is premium

Preview:
image

PR #116, opened 2021-08-24
Add Premiun Panel State to Add On V2

codemist:
This PR implements the new premium add on design:

- Adds dashboard data from profile page into add on
- Add CTA to get custom domain from add on
- When user is premium, switch completely to a new panel that displays premium content

WIP:
- Switching from the get custom domain CTA + illustration module to the educational module if a user has a custom domain

Notes: 
- Waiting on 'get email illustration' to get greenlighted by UX 
- Revised version of old PR https://github.com/mozilla/fx-private-relay-add-on/pull/114
-


src/js/popup.js
```diff
@@ -54,9 +90,27 @@ async function showRelayPanel(tipPanelToShow) {
...
+    //Premium Panel
+    registerDomainImgEl.src = panelStrings.registerDomainImg;
+    registerDomainButtonEl.textContent = panelStrings.registerDomainButton;
+    registerDomainHeadlineEl.textContent = panelStrings.registerDomainHeadline;
+    aliasesUsedValEl.textContent = aliasesUsedVal;
+    emailsBlockedValEl.textContent = emailsBlockedVal;
+    emailsForwardedValEl.textContent = emailsForwardedVal;
+    aliasesUsedTextEl.textContent = panelStrings.aliasesUsedText;
+    emailsBlockedTextEl.textContent = panelStrings.emailsBlockedText;
+    emailsForwardedTextEl.textContent = panelStrings.emailsForwardedText;
```
groovecoder, line 103: There's code in the [i18n.js](https://github.com/mozilla/fx-private-relay-add-on/blob/main/src/js/i18n.js) script that looks at every element with `class="i18n-content"`, and sets the `textContent` to the translated message using the message id value from the element's `data-i18n-message-id` attribute.

So, as long as all these elements have `class=i18n-content` and the right `data-i18n-message-id` attribute, you shouldn't need to assign the `textContent` value yourself.

groovecoder: And once you remove these, you can probably remove all of the `premiumPanelWrapper.querySelector` lines above that are only getting the element in order to set the `textContent`.

groovecoder (changes requested):
Just about done ... the main thing is to make use of the `i18n.js` functionality to help yourself not have to repeat so many JS lines. 😉

src/js/popup.js
codemist: Oh nice, I was wondering why some tags had that `data-i18n-message-id` attribute while some didn't. This makes sense to me now, thank you!

src/popup.html
```diff
@@ -58,28 +58,49 @@ <h1 class="settings-hl i18n-content" data-i18n-message-id="popupSettingsTitle"><
...
+                </li>
+                <li class="dashboard-info">
+                  <span class="emails-forwarded-text i18n-content" data-i18n-message-id="popupEmailsForwarded"></span>
+                  <div class="dashboard-stats emails-forwarded"></div>
+                </li>
+              </ul>
+            </div>
+            <img class="email-domain-illustration">
+            <p class="register-domain-headline i18n-content" data-i18n-message-id="popupRegisterDomainHeadline"></p>
+            <a class="register-domain-cta i18n-content button sign-in-btn blue-primary-btn close-popup-after-click" data-event-action="click" data-i18n-message-id="registerDomainButton"></a>
```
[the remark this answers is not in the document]
codemist: Ahhh yes classic mistake, thanks for the catch

PR #130, opened 2021-09-02
Fix panel bar l10n strings

codemist:
Some of the l10n strings for the panel bar on non premium view were mis-IDed. Should work now: [demo](https://drive.google.com/file/d/1E6M4JuAuCajdPDOi2DNxRJXmPEu_9Q18/view?usp=sharing)


groovecoder (approved):
ooo, thanks for fixing the things I broke. sorry!

PR #131, opened 2021-09-02
Fix subdomain module switching logic

codemist:
This Pr uses `is-hidden` instead of removing the modules completely, and checks if user has a subdomain data as a string.

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -66,11 +66,16 @@ function choosePanel(numRemaining, panelId, premium){
 }
 
 function checkUserSubdomain(premiumSubdomainSet){
-  if(premiumSubdomainSet != "None"){
-    document.getElementsByClassName("register-domain-component")[0].remove();
+  const educationalComponent = document.querySelector(".educational-component");
+  const registerDomainComponent = document.querySelector(".register-domain-component");
+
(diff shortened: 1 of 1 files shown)


src/js/popup.js
```diff
@@ -131,6 +133,11 @@ async function showRelayPanel(tipPanelToShow) {
   const { emailsBlockedVal } = await browser.storage.local.get("emailsBlockedVal");
 
 
+  //Subdomain Data
+  const { premiumSubdomainSetString } = await browser.storage.local.get("premiumSubdomainSet");
+  checkUserSubdomain(premiumSubdomainSetString);
```
groovecoder, line 138: Oops, I may have given (MOAR!) bad advice here. I think the de-structuring assignment syntax wants the variable name to be the exact same name as the object property. 😢 

So I think it has to be:
```suggestion
  const {premiumSubdomainSet} = await browser.storage.local.get("premiumSubdomainSet");
  checkUserSubdomain(premiumSubdomainSet);
```

groovecoder (changes requested):

src/js/popup.js
codemist: Done! I added spaces around `{ premiumSubdomainSet }`, hope that's ok!

PR #132, opened 2021-09-02
Add Get Unlimited Aliases CTA on Field Input Popup and Context Menu

codemist:
UPDATED DEMO: https://drive.google.com/file/d/12LaQTdpuV9Vfx_2d_OH_2Eir2o8FIBR9/view?usp=sharing

Note: Signing in and out after upgrading to activate premium because I'm using my local server to test - FXA sends its "subscription change event" webhooks to the Relay server url (from @groovecoder). 

This PR should
1) Remove any CTA to upgrade on the context menu/field input popup when `PREMIUM_ENABLED` is set to `False` or when user is already premium
2) Show CTA to upgrade for free users on the context menu/field input popup if `PREMIUM_ENABLED` is set to `True`.
3) Remove `'Generate new alias'` on the field input popup if free user has exhausted all 5 free aliases.

### DIFF
--- a/src/css/relay.css
+++ b/src/css/relay.css
@@ -199,15 +199,13 @@ button.fx-relay-modal-close-button {
   flex-direction: column;
   position: absolute;
   align-content: center;
-  padding-top: 20px;
-  padding-left: 0;
-  padding-right: 0;
-  padding-bottom: 0;
+  padding: 24px 16px 16px 16px;
   top: 47px;
   right: -15px;
   pointer-events: all;
   animation: fxRelayFadeIn 0.2s ease forwards;
   box-shadow: 3px 3px 3px 3px #00000017;
+  box-sizing: border-box;
 }
 
 .fx-relay-menu::before {
@@ -242,74 +240,82 @@ button.fx-relay-modal-close-button {
 
 .fx-relay-menu-remaining-aliases {
   font-size: 14px;
-  width: 260px;
-  padding: 12px 0;
+  width: 100%;
+  padding: 16px 0px;
   background: var(--relayGrey20);
   color: var(--relayGrey60);
   text-align: center;
-  font-weight: 600;
-  font-size: 14px;
   border-radius: 4px;
-  margin: auto auto 12px auto;
+  margin-bottom: 8px;
+  font-weight: 600;
 }
 
-button.fx-relay-menu-sign-up-btn,
-.fx-relay-menu-dashboard-link {
+button.fx-relay-menu-sign-up-btn{
   font-size: 15px;
   border: 1px solid rgba(0, 0, 0, 0);
   border-color: rgba(0, 0, 0, 0);
   outline: none;
   width: 100%;
   min-width: 100%;
-  font-family: var(--relayMetropolis) !important;
-  color: var(--relayInk) !important;
+  font-family: var(--relayMetropolis);
+  color: var(--relayInk);
   display: flex;
   min-height: auto;
   box-shadow: none;
   box-sizing: border-box;
-  background-color: var(--relayGrey20) !important;
+  background-color: var(--relayGrey20);
   text-align: center;
   justify-content: center;
   padding: 12px 24px;
   pointer-events: all;
   border-bottom-left-radius: 4px;
   border-bottom-right-radius: 4px;
-  text-decoration: none !important;
+  text-decoration: none;
 }
 
 .fx-relay-menu-dashboard-link {
-  margin: auto auto 0 auto;
+  color: var(--relayBlue3);
+  font-size: 15px;
+  font-family: "InterUI", sans-serif;
+  text-align: center;
+  padding: 8px;
+  margin: 16px 0 0 0;
+  border-radius: 4px;
+}
+
+.fx-relay-menu-dashboard-link:hover {
+  background: var(--relayGrey20);
+  text-decoration: none;
+  color: var(--relayBlue3);
 }
 
 button.fx-relay-menu-generate-alias-btn {
   background-color: var(--relayBlue3) !important;
-  padding: 12px !important;
-  width: 260px !important;
+  width: 100%;
+  padding: 16px 0px;
   border-radius: 4px !important;
   font-family: var(--relayMetropolis) !important;
   color: rgba(255, 255, 255, 1) !important;
   border: none !important;
   font-size: 16px !important;
-  margin: auto auto 16px auto !important;
   text-align: center !important;
-  font-weight: 500 !important;
+  margin-top: 4px;
 }
 
 .fx-relay-menu-get-unlimited-aliases {
-  border: 2px solid var(--relayBlue3) !important;
-  padding: 12px !important;
-  width: 240px !important;
-  border-radius: 4px !important;
-  font-family: var(--InterUI) !important;
-  color: var(--relayBlue3) !important;
-  font-size: 16px !important;
-  text-align: center !important;
-  font-weight: 500 !important;
-  margin: auto auto 16px auto !important;
+  border: 2px solid var(--relayBlue3);
+  width: 100%;
+  padding: 16px 0px;
+  border-radius: 4px;
+  font-family: "InterUI", sans-serif;
+  color: var(--relayBlue3);
+  font-size: 16px;
+  text-align: center;
 }
 
 .fx-relay-menu-get-unlimited-aliases:hover {
   text-decoration: none;
+  color: white;
   background-color: var(--relayBlueHover) !important;
 }
 
@@ -328,7 +334,6 @@ button.fx-relay-menu-generate-alias-btn[disabled] {
 }
 
 .fx-relay-menu-sign-up-btn:hover,
-.fx-relay-menu-dashboard-link:hover,
 .fx-relay-modal-close-button:hover {
   background-color: var(--relayGrey30) !important;
 }
--- a/src/css/relay.css
+++ b/src/css/relay.css
@@ -247,7 +247,16 @@ button.fx-relay-modal-close-button {
   text-align: center;
   margin-top: 8px;
   margin-bottom: 24px;
-  font-weight: 600 !important;
+  font-weight: 600;
+}
+
+.max-num-aliases {
+  background: var(--relayGrey30);
+  color: var(--relayGrey60);
+  padding: 16px;
+  font-size: 14px;
+  border-radius: 4px;
+  margin: 12px 16px;
 }
 
 button.fx-relay-menu-sign-up-btn,
@@ -293,13 +302,12 @@ button.fx-relay-menu-generate-alias-btn {
 }
 
 .fx-relay-menu-get-unlimited-aliases {
-  background-color: var(--relayBlue3) !important;
+  border: 2px solid var(--relayBlue3) !important;
   padding: 12px !important;
-  width: 260px !important;
+  width: 240px !important;
   border-radius: 4px !important;
-  font-family: var(--relayMetropolis) !important;
-  color: rgba(255, 255, 255, 1) !important;
-  border: none !important;
+  font-family: var(--InterUI) !important;
+  color: var(--relayBlue3) !important;
   font-size: 16px !important;
   margin: auto !important;
   text-align: center !important;
--- a/src/css/relay.css
+++ b/src/css/relay.css
@@ -289,7 +289,7 @@ button.fx-relay-menu-generate-alias-btn {
 
 .fx-relay-menu-get-unlimited-aliases:hover {
   text-decoration: none;
-  color: white;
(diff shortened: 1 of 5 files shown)


src/js/add_input_icon.js
```diff
@@ -240,14 +238,31 @@ async function addRelayIconToInput(emailInput) {
       sendInPageEvent("click", "input-menu-manage-all-aliases-btn");
     });
 
+    //Create "Get unlimited aliases" link
+    const { fxaSubscriptionsUrl } = await browser.storage.local.get("fxaSubscriptionsUrl");
+    const { premiumProdId } = await browser.storage.local.get("premiumProdId");
+    const { premiumPriceId } = await browser.storage.local.get("premiumPriceId");
```
groovecoder, line 244: `browser.storage.local.get` will let you pass an array of strings so you can make just 1 call: `browser.storage.local.get(["fxaSubscriptionsUrl", "premiumProdId", "premiumPriceId"]);`

Looks like that will return a results object containing all 3 objects inside it.

src/js/add_input_icon.js
```diff
@@ -240,14 +238,31 @@ async function addRelayIconToInput(emailInput) {
...
-    [generateAliasBtn, remainingAliasesSpan, relayMenuDashboardLink].forEach(el => {
+    [getUnlimitedAliasesBtn, generateAliasBtn, remainingAliasesSpan, relayMenuDashboardLink].forEach(el => {
       relayInPageMenu.appendChild(el);
     });
 
+    //Show get unlimited aliases btn
+    if (!premium) {
+      if (maxNumAliasesReached) {
+        generateAliasBtn.remove();
+        sendInPageEvent("viewed-menu", "input-menu-max-aliases-message")
```
groovecoder, line 259: nice telemetry ping!

src/js/add_input_icon.js
codemist: Should I do `const { fxaSubscriptionsUrl, premiumProdId, premiumPriceId } = await browser.storage.local.get("fxaSubscriptionsUrl", "premiumProdId", "premiumPriceId");
` ?

codemist:
Note: may be subject to change after Eduardo's context menu design demo on Sept 9.

maxxcrawford (changes requested):
The hover state is blue-on-blue: 

image

src/css/relay.css
```diff
@@ -242,12 +242,15 @@ button.fx-relay-modal-close-button {
 
 .fx-relay-menu-remaining-aliases {
   font-size: 14px;
-  padding: 0 24px;
-  color: var(--relayInk70);
+  width: 260px;
```
maxxcrawford, line 245: Is the hard-width here required? It feels like a magic number and could break funny in different languages.

src/css/relay.css
```diff
@@ -242,12 +242,15 @@ button.fx-relay-modal-close-button {
...
+  padding: 12px 0;
+  background: var(--relayGrey20);
+  color: var(--relayGrey60);
   text-align: center;
-  margin-top: 8px;
-  margin-bottom: 24px;
-  font-weight: 600 !important;
+  font-weight: 600;
+  font-size: 14px;
```
maxxcrawford, line 251: `font-size` is listed twice in this rule:
```suggestion
```

src/css/relay.css
```diff
@@ -242,12 +242,15 @@ button.fx-relay-modal-close-button {
...
+  border-radius: 4px;
+  margin: auto auto 12px auto;
```
maxxcrawford, line 253: I would recommend sticking with a [Protocol token](https://protocol.mozilla.org/fundamentals/tokens.html) if possible (`8px`/`16px`) versus `12px`;

src/css/relay.css
```diff
@@ -242,12 +242,15 @@ button.fx-relay-modal-close-button {
...
-  color: var(--relayInk70);
+  width: 260px;
+  padding: 12px 0;
+  background: var(--relayGrey20);
+  color: var(--relayGrey60);
   text-align: center;
-  margin-top: 8px;
-  margin-bottom: 24px;
-  font-weight: 600 !important;
+  font-weight: 600;
```
maxxcrawford, line 250: This font-weight seems overly specific (_I'm expecting bold or not bold_). Do we have that specific font-weight available?

src/css/relay.css
```diff
@@ -280,22 +283,36 @@ button.fx-relay-menu-sign-up-btn,
...
   background-color: var(--relayBlue3) !important;
-  padding-top: 12px !important;
-  padding-bottom: 12px !important;
-  padding-left: 28px !important;
-  padding-right: 28px !important;
-  max-width: 260px;
-  width: 100%;
+  padding: 12px !important;
+  width: 260px !important;
+  border-radius: 4px !important;
```
maxxcrawford, line 288: Do we need the `important!` flags?

src/css/relay.css
```diff
@@ -280,22 +283,36 @@ button.fx-relay-menu-sign-up-btn,
...
-  margin: auto !important;
+  margin: auto auto 16px auto !important;
   text-align: center !important;
   font-weight: 500 !important;
 }
 
+.fx-relay-menu-get-unlimited-aliases {
+  border: 2px solid var(--relayBlue3) !important;
+  padding: 12px !important;
+  width: 240px !important;
```
maxxcrawford, line 301: Another `width` property. I would recommend removing, if possible.

src/css/relay.css
```diff
@@ -280,22 +283,36 @@ button.fx-relay-menu-sign-up-btn,
...
+  border-radius: 4px !important;
+  font-family: var(--InterUI) !important;
+  color: var(--relayBlue3) !important;
+  font-size: 16px !important;
+  text-align: center !important;
+  font-weight: 500 !important;
```
maxxcrawford, line 307: Same `font-weight` question from above. Why `600` there and `500` here?

src/css/relay.css
maxxcrawford: Checking here: https://github.com/mozilla/fx-private-relay-add-on/blob/main/src/css/relay.css#L1-L26

src/css/relay.css
```diff
@@ -280,22 +283,36 @@ button.fx-relay-menu-sign-up-btn,
...
+  margin: auto auto 16px auto !important;
+}
+
+.fx-relay-menu-get-unlimited-aliases:hover {
+  text-decoration: none;
+  background-color: var(--relayBlueHover) !important;
+}
```
maxxcrawford, line 314: This may fix the hover text-color issue. 

```suggestion
.fx-relay-menu-get-unlimited-aliases:hover {
  text-decoration: none;
  background-color: var(--relayBlueHover) !important;
  color: white
}
```

src/js/add_input_icon.js
```diff
@@ -240,14 +238,44 @@ async function addRelayIconToInput(emailInput) {
...
+    else {
+      getUnlimitedAliasesBtn.remove();
+    }
+
+    const premiumEnabled = await browser.storage.local.get("premiumEnabled");
+    const premiumEnabledString = premiumEnabled.premiumEnabled;
+
+    if(premiumEnabledString != "True") {
+      getUnlimitedAliasesBtn.remove();
+    }
```
maxxcrawford, line 277: I feel like you could add this if statement to the above main check and chain it with some or `||` logic.

maxxcrawford (changes requested):
I'm using a premium test account and still get the upgrade prompt: 

image

groovecoder:
Oops, this needs some conflict resolution.

src/js/add_input_icon.js
```diff
@@ -216,6 +224,9 @@ async function addRelayIconToInput(emailInput) {
     const generateAliasBtn = createElementWithClassList("button", "fx-relay-menu-generate-alias-btn");
     generateAliasBtn.textContent = browser.i18n.getMessage("pageInputIconGenerateNewAlias");
 
+    // Create "Get unlimited aliases" button
+    const getUnlimitedAliasesBtn = createElementWithClassList("a", "fx-relay-menu-get-unlimited-aliases");
+    getUnlimitedAliasesBtn.textContent = browser.i18n.getMessage("popupGetUnlimitedAliases");
```
Vinnl, line 229: We might want to avoid replacing the user's tab:

```suggestion
    getUnlimitedAliasesBtn.textContent = browser.i18n.getMessage("popupGetUnlimitedAliases");
    getUnlimitedAliasesBtn.setAttribute("target", "_blank");
    // See https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/
    getUnlimitedAliasesBtn.setAttribute("rel", "noopener noreferrer");
```

Vinnl (approved):
Obviously not super familiar with the code base, but looks good to me! Left some minor suggestions for your consideration inline.

Also, I'm not sure, but I don't think there should be changes in `src/_locales`? (i.e. after submitting locale changes to [the l10n repo](https://github.com/mozilla-l10n/fx-private-relay-add-on-l10n), they should make it to this repo automatically?)

src/css/relay.css
```diff
@@ -311,7 +334,6 @@ button.fx-relay-menu-generate-alias-btn[disabled] {
 }
 
 .fx-relay-menu-sign-up-btn:hover,
-.fx-relay-menu-dashboard-link:hover,
```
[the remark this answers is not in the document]
maxxcrawford: I went and double-checked. There's still `:hover` styles being applied to `.fx-relay-menu-dashboard-link` here: https://github.com/mozilla/fx-private-relay-add-on/blob/get-unlimited-cta/src/css/relay.css#L286-L290

@codemist Is that style enough?

maxxcrawford (changes requested):
I'm now having the opposite issue from [my previous comment](https://github.com/mozilla/fx-private-relay-add-on/pull/132#pullrequestreview-756522514) – if I'm in a premium account, I do not get the upgrade context menu item, as expected. However, if I'm NOT premium, I do not get the prompt either (when I should).

Right+click as a non-premium account. I do, however, get the correct prompts in the popup/panel.

image

image

src/css/relay.css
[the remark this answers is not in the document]
codemist: Yeah, this was following the new design spec which is different from `.fx-relay-menu-sign-up-btn:hover`

src/js/add_input_icon.js
```diff
@@ -227,6 +227,8 @@ async function addRelayIconToInput(emailInput) {
     // Create "Get unlimited aliases" button
     const getUnlimitedAliasesBtn = createElementWithClassList("a", "fx-relay-menu-get-unlimited-aliases");
     getUnlimitedAliasesBtn.textContent = browser.i18n.getMessage("popupGetUnlimitedAliases");
+    getUnlimitedAliasesBtn.setAttribute("target", "_blank");
+    getUnlimitedAliasesBtn.setAttribute("rel", "noopener noreferrer");
```
Vinnl, line 231: Since it's not super obvious why this attribute needs to be added, it could be helpful for future readers to add a link to an explanatory article, e.g. https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/.

Vinnl (approved):

codemist:
@Vinnl @maxxcrawford Updated the link to the demo at the top, walkthrough of the feature from free to premium account.

codemist:
@Vinnl @maxxcrawford One last check?

src/js/get_profile_data.js
```diff
@@ -178,4 +178,10 @@
     relayAddresses.push(relayAddress);
   }
   browser.storage.local.set({relayAddresses});
+
+  console.log("get_profile_data--sendMessage");
```
Vinnl, line 182: Leftover from debugging?

```suggestion
```

Vinnl (commented):
Just gave it another try locally, two things I noticed:

- Clicking "Get unlimited aliases" in the menu on the Relay icon in the email field does not open in a new tab.
- Clicking "Generate New Alias" in the context menu when I've reached the limit does not prompt me to upgrade. From your video it looks like it should've been? (Edit: ha, it seems my right-click menu doesn't get caught by my screen record, in case it's confusing :P )

https://user-images.githubusercontent.com/4251/136008871-62c34ead-f145-4b8e-811f-66cf03e8e4e9.mp4

codemist:
@Vinnl That's strange. Did you pull the recent changes? My prompt looks like this for me: 
image

Vinnl:
Aargh sorry about that, hadn't pulled the latest changes. Yes, I do get the correct prompt now - only very minor quibble is that the hover style makes the text unreadable:

image

maxxcrawford (approved):
A month of hard work. Great, great, great job, @codemist! 👏

PR #133, opened 2021-09-03
Fix #118 - Only show premium UI if premium is enabled [NEW VERSION]

### LINKED ISSUE #118: Only show premium/upgrade CTAs if Relay website has premium enabled.
Affected areas: 
- Popup
- Content Script
- Context menu 

## Acceptance Criteria 

- [ ] Current (production) UI should be maintained as default (So everything new is hidden by default
- [ ] Only show upgrade CTAs across the add-on if the Relay website it connects to has premium enabled

codemist:
[REVISED PR]

This PR:

- Takes into account the value of `settings.PREMIUM_ENABLED`, which checks whether premium features are available on the site.

- So if the aforementioned value doesn't return True, the add on defaults back to a 'non-premium' view.

- All premium UI blocks now have a `is-hidden` utility class.

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -51,11 +51,39 @@ function showSignUpPanel() {
 }
 
 
-function choosePanel(numRemaining, panelId, premium, premiumEnabledString){
-  if (premium && premiumEnabledString === "True"){
+function premiumFeaturesAvailable(premiumEnabledString) {
+  if (premiumEnabledString === "True") {
+    return Boolean(true);
+  }
+  else {
+    return Boolean(false);
+  }
+}
+
+function enablePremium(premium, premiumEnabledString) {
+  if (premium && premiumFeaturesAvailable(premiumEnabledString)) {
+    return Boolean(true);
+  }
+  else {
+    return Boolean(false);
+  }
+}
+
+
+function choosePanel(numRemaining, panelId, premium, premiumEnabledString, premiumSubdomainSet){
+  const premiumPanelWrapper = document.querySelector(".premium-wrapper");
+
+  if (enablePremium(premium, premiumEnabledString)){
     document.getElementsByClassName("content-wrapper")[0].remove();
+    premiumPanelWrapper.classList.remove("is-hidden");
+    premiumPanelWrapper.querySelectorAll(".is-hidden").forEach(premiumFeature => 
+      premiumFeature.classList.remove("is-hidden") 
+      );
+    //Toggle register domain or education module
+    checkUserSubdomain(premiumSubdomainSet);
     return 'premiumPanel';
   }
+
   else {
     const premiumWrapper = document.getElementsByClassName("premium-wrapper")
     if (premiumWrapper.length) {
@@ -102,9 +130,13 @@ async function showRelayPanel(tipPanelToShow) {
 
   //Check if user is premium
   const { premium } = await browser.storage.local.get("premium");
+  
+  //Check if user has a subdomain set
+  const { premiumSubdomainSet } = await browser.storage.local.get("premiumSubdomainSet");
+
 
   const updatePanel = (numRemaining, panelId) => {
-    const panelToShow = choosePanel(numRemaining, panelId, premium, premiumEnabledString);
+    const panelToShow = choosePanel(numRemaining, panelId, premium, premiumEnabledString, premiumSubdomainSet);
     onboardingPanelWrapper.classList = [panelToShow];
     const panelStrings = onboardingPanelStrings[`${panelToShow}`];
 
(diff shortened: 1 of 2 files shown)


src/js/popup.js
```diff
@@ -117,6 +121,15 @@ async function showRelayPanel(tipPanelToShow) {
     emailsBlockedValEl.textContent = emailsBlockedVal;
     emailsForwardedValEl.textContent = emailsForwardedVal;
 
+    //Show premium panel state
+    if (premium && premiumEnabledString === "True"){
```
codemist, line 125: This if statement un-hides premium UI blocks if the user is premium and if premium features are available.

src/js/popup.js
```diff
@@ -51,8 +51,8 @@ function showSignUpPanel() {
 }
 
 
-function choosePanel(numRemaining, panelId, premium){
-  if (premium){
+function choosePanel(numRemaining, panelId, premium, premiumEnabledString){
+  if (premium && premiumEnabledString === "True"){
```
codemist, line 55: Passing in the string value of `settings.PREMIUM_ENABLED`, so the premium panel is set only if both the user data attribute is premium and if the premium features are available on the site.



And if `premiumEnabledString` doesn't return `True`, the add on defaults back to a 'non-premium' view.

src/popup.html
```diff
@@ -77,8 +77,9 @@ <h1 class="onboarding-h1"></h1>
             </div>
           </div>
           <!-- Premium Panel-->
-          <div class="premium-wrapper">
-            <div class="premium-dashboard">
+          <div class="premium-wrapper is-hidden">
```
codemist, line 80: Added a `is-hidden` utility class to every premium UI block/module.

src/js/popup.js
```diff
@@ -117,6 +121,15 @@ async function showRelayPanel(tipPanelToShow) {
     emailsBlockedValEl.textContent = emailsBlockedVal;
     emailsForwardedValEl.textContent = emailsForwardedVal;
 
+    //Show premium panel state
+    if (premium && premiumEnabledString === "True"){
+      premiumPanelWrapper.classList.remove("is-hidden");
+      premiumPanelWrapper.querySelectorAll(".is-hidden").forEach(premiumFeature => 
+        premiumFeature.classList.remove("is-hidden") 
+        );
```
codemist, line 129: Checks all children of the `premium-wrapper` and removes their `is-hidden` class.

src/js/popup.js
```diff
@@ -117,6 +121,15 @@ async function showRelayPanel(tipPanelToShow) {
...
+      //Toggle register domain or education module
+      checkUserSubdomain(premiumSubdomainSet);
```
codemist, line 131: After checking if user is premium, this line goes on to check which module to show depending on whether a subdomain is set or not.

codemist: Since all the premium UI blocks are set to visible when they're premium, depending on whether or not a user has a subdomain, this function will determine which module to display by hiding the opposing one (register domain CTA vs educational module).

codemist (commented):
Started a self review

maxxcrawford (changes requested):
I see one upgrade item visible when it shouldn't be: 

image

src/js/popup.js
```diff
@@ -92,15 +92,19 @@ async function showRelayPanel(tipPanelToShow) {
...
   const premiumPanelWrapper = document.querySelector(".premium-wrapper");
   const registerDomainImgEl = premiumPanelWrapper.querySelector(".email-domain-illustration");
-
   const aliasesUsedValEl = premiumPanelWrapper.querySelector(".aliases-used");
   const emailsBlockedValEl = premiumPanelWrapper.querySelector(".emails-blocked");
   const emailsForwardedValEl = premiumPanelWrapper.querySelector(".emails-forwarded");
 
+  //Check if premium features are available
+  const premiumEnabled = await browser.storage.local.get("premiumEnabled");
+  const premiumEnabledString = premiumEnabled.premiumEnabled;
```
pdehaan, line 101: I _think_ this is the only place where we used `premiumEnabled`. Not sure if you can just destructure this and reduce by 1 line.

```js
const { premiumEnabled } = await browser.storage.local.get("premiumEnabled");
```

pdehaan: Or you could even wrap this in some helper function which also converts from "True" string to a Boolean `true`.

codemist: I remember @groovecoder and I tried to destructure lines L100 and L101 but for some reason it would output `undefined` when we tried to console log it.

maxxcrawford (approved):
Approved pending fixes from @pdehaan! Everything looks/works as expected.

src/js/popup.js
```diff
@@ -51,11 +51,39 @@ function showSignUpPanel() {
...
+function premiumFeaturesAvailable(premiumEnabledString) {
+  if (premiumEnabledString === "True") {
+    return Boolean(true);
+  }
+  else {
+    return Boolean(false);
+  }
+}
+
+function enablePremium(premium, premiumEnabledString) {
```
codemist, line 63: This is function is to check if 1) user is premium 2) premium features are available. Different from the earlier function which only checks the `premiumEnabledString` string, since I have to use that for a specific case.

src/js/popup.js
```diff
@@ -51,11 +51,39 @@ function showSignUpPanel() {
...
+  else {
+    return Boolean(false);
+  }
+}
+
+
+function choosePanel(numRemaining, panelId, premium, premiumEnabledString, premiumSubdomainSet){
+  const premiumPanelWrapper = document.querySelector(".premium-wrapper");
+
+  if (enablePremium(premium, premiumEnabledString)){
```
codemist, line 76: Grouped up all of the features that meet criteria 1) and 2) into an if-statement for tidiness.

1 ) user is premium 2) premium features

src/js/popup.js
```diff
@@ -115,21 +147,27 @@ async function showRelayPanel(tipPanelToShow) {
...
+    //If Premium features are not available, do not show upgrade CTA on the panel
+    if (!premiumFeaturesAvailable(premiumEnabledString)) {
+      const premiumCTA = document.querySelector(".premium-cta");
+      const upgradeBanner = document.querySelector(".upgrade-banner-wrapper");
+      premiumCTA.classList.add("is-hidden");
+      upgradeBanner.classList.add("is-hidden");
+    }
+
+    // Remove panel status if user has unlimited aliases, so no negative alias left count
+    if (premium) {
```
codemist, line 166: This hides a specific UI element so long as a user has unlimited alias so there would be no negative alias left count shown.

src/js/popup.js
```diff
@@ -51,11 +51,39 @@ function showSignUpPanel() {
...
+    return Boolean(true);
+  }
+  else {
+    return Boolean(false);
+  }
+}
+
+function enablePremium(premium, premiumEnabledString) {
+  if (premium && premiumFeaturesAvailable(premiumEnabledString)) {
+    return Boolean(true);
```
pdehaan, line 65: Not sure you need to wrap in `Boolean()`. You _could_ also reduce this to the following (although not convinced it's better/legible):

```js
return premium && premiumFeaturesAvailable(premiumEnabledString);
```

src/js/popup.js
```diff
@@ -51,11 +51,39 @@ function showSignUpPanel() {
 }
 
 
-function choosePanel(numRemaining, panelId, premium){
-  if (premium){
+function premiumFeaturesAvailable(premiumEnabledString) {
+  if (premiumEnabledString === "True") {
+    return Boolean(true);
```
pdehaan, line 56: Since you're using `return` in the `if`, you can drop the `else`:

```js
  if (premiumEnabledString === "True") {
    return true;
  }
  return false;
```

Or one-liner it:

```js
  return premiumEnabledString === "True";
```

maxxcrawford (approved):
This works as expected. Thanks!

PR #140, opened 2021-09-15
Preserve input width of content-box-sized elements

### LINKED ISSUE #103: Email field issue on some sites when relay add-on is installed and relay icon activated (facebook)
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit

**Browsers tested on:**
Firefox

**Preconditions:**
The user is logged in to Firefox Relay page.
Relay add-on is installed.
The user is logged out of Facebook.

**Steps to reproduce:**

1. Navigate to https://www.facebook.com/
2. Observe the email field.

**Actual result:**
The email field gets outside of the registration form when the relay icon is displayed.

**Expected result:** 
The email field is correctly displayed.

facebook relay icon issue

Vinnl:
Fixes #103.

Inputs whose box-sizing was not border-box, such as the one on the
Facebook homepage, would increase in size due to the added padding.
Additionally, if existing padding was present on the input, that
would be disregarded when making space for the Relay icon.

Both should now work correctly.

I also reduced the padding that's added to be equivalent to the Relay icon width, since now that the original padding is preserved that should leave enough space for the input text not to be nudged against the icon.

### DIFF
--- a/src/js/add_input_icon.js
+++ b/src/js/add_input_icon.js
@@ -99,6 +99,23 @@ async function isUserSignedIn() {
   return (userApiToken.hasOwnProperty("apiToken"));
 }
 
+function addPaddingRight(element, paddingInPixels) {
+  const computedElementStyles = getComputedStyle(element);
+  const existingPaddingRight = computedElementStyles.getPropertyValue("padding-right");
+  const existingPaddingRightInPixels = Number.parseInt(existingPaddingRight.replace("px", ""), 10);
+  const newPaddingRight = existingPaddingRightInPixels + paddingInPixels;
+  element.style.paddingRight = newPaddingRight.toString() + "px";
+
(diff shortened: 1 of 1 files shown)


maxxcrawford (approved):
I checked on Facebook.com, Airbnb.com and Imgur.com. All worked as expected: 

Facebook: 
image

Imgur:
image

Airbnb has some padding that exists, but I think this is correct? Please advise. 
image

Vinnl:
This is what AirBnb looks like for me without this change, so while it doesn't look great, I think it's a separate issue.

image

maxxcrawford:
> @maxxcrawford This is what AirBnb looks like for me without this change, so while it doesn't look great, I think it's a separate issue.
> 

Agreed. Here's what it looks on the current published release: 

image

Merging this in!

PR #141, opened 2021-09-15
Store the original site when generating a new alias from the add-on

maxxcrawford:
This PR adds a new data point to aliases generated from the add-on. This data attribute will be used in a later release to recommend labels to be reused on sites that match the same domain

Note that this data is only stored locally for the user. In a future release, we will use this information to display an already generated alias on sites that match the site origin data attribute. 

This PR fixes MPP-847 

## Testing 

- To verify, generate a new alias using the content script/context menu from a website
- Debug the add-on and run the following command in the console. This will return an array of all your created aliases. Look at the alias you just created: 

  ` await browser.storage.local.get("relayAddresses")`

- **Expected:**  The first item in the array (newest created alias) should contain the root domain (same value that matches the inital label value) in the `siteOrigin` property. 
- Edit the label to something else. Re-run the console command.
- **Expected: ** The array should now return the revised label with the siteOrigin unchanged.

### DIFF
--- a/src/js/get_profile_data.js
+++ b/src/js/get_profile_data.js
@@ -45,6 +45,10 @@
 
     const defaultAliasLabelText = browser.i18n.getMessage("profilePageDefaulAliasLabelText");
     const storedAliasLabel = (addonRelayAddress && addonRelayAddress.hasOwnProperty("domain")) ? addonRelayAddress.domain : "";
+    
+    // Cache the siteOrigin alias attribute when updating local storage data.
+    // Note that this data attribute only exists in aliases generated through the add-on
+    const storedAliasSiteOrigin = (addonRelayAddress && addonRelayAddress.hasOwnProperty("siteOrigin")) ? addonRelayAddress.siteOrigin : "";
 
     const aliasLabelForm = aliasCard.querySelector("form.relay-email-address-label-form");
     const aliasLabelInput = aliasCard.querySelector("input.relay-email-address-label");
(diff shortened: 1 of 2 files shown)


src/js/get_profile_data.js
```diff
@@ -45,6 +45,10 @@
 
     const defaultAliasLabelText = browser.i18n.getMessage("profilePageDefaulAliasLabelText");
     const storedAliasLabel = (addonRelayAddress && addonRelayAddress.hasOwnProperty("domain")) ? addonRelayAddress.domain : "";
+    
+    // Cache the siteOrigin alias attribute when updating local storage data.
+    // Note that this data attribute only exists in aliases generated through the add-on
+    const storedAliasSiteOrigin = (addonRelayAddress && addonRelayAddress.hasOwnProperty("siteOrigin")) ? addonRelayAddress.siteOrigin : "";
```
maxxcrawford, line 51: As this data only exists from the add-on/local storage, it has to be queried the same was as the alias label is.

Vinnl: Some small syntactic sugar you might be interested in :)

```suggestion
    
    // Cache the siteOrigin alias attribute when updating local storage data.
    // Note that this data attribute only exists in aliases generated through the add-on
    const storedAliasSiteOrigin = addonRelayAddress?.siteOrigin ?? "";
```

Vinnl (approved):
Works as expected here.

PR #144, opened 2021-09-20
Use an inline wrapper for input fields

Vinnl:
Fixes # #143.

The input field at https://www.cookson-clal.com/login_in.jsp?country=gb&shop=c&language=FRA&site400=001&company=004&siteid=IC is not clickable - our `<div>` wrapper around the input is inside a `<span>` wrapper by the site. As far as I can see there's no need for our wrapper to be a block-level element.

### DIFF
--- a/src/css/relay.css
+++ b/src/css/relay.css
@@ -348,6 +348,7 @@ button.fx-relay-menu-generate-alias-btn[disabled] {
 }
 
 .fx-relay-email-input-wrapper {
+  display: inline-block;
   position: relative;
   pointer-events: none;
   padding: 0 !important;
--- a/src/js/other-websites/add_input_icon.js
+++ b/src/js/other-websites/add_input_icon.js
@@ -135,7 +135,7 @@ async function addRelayIconToInput(emailInput) {
 
   // create new wrapping element;
   const emailInputWrapper = createElementWithClassList(
-    "div",
+    "span",
     "fx-relay-email-input-wrapper"
   );
   emailInputOriginalParentEl.insertBefore(emailInputWrapper, emailInput);


src/js/add_input_icon.js
```diff
@@ -123,7 +123,7 @@ async function addRelayIconToInput(emailInput) {
   const emailInputOriginalParentEl = emailInput.parentElement;
 
   // create new wrapping element;
-  const emailInputWrapper = createElementWithClassList("div", "fx-relay-email-input-wrapper");
+  const emailInputWrapper = createElementWithClassList("span", "fx-relay-email-input-wrapper");
```
maxxcrawford, line 126: Would it be possible to use a custom element (via `customElements.define…`?) and avoid inherited styles? I say that bc this change will absolutely fix Issue #1114, but it may open us up to entirely new website issues who have styles applied to `spans` in their email field/login areas.

maxxcrawford (approved):
I'm tenativey approving this, but let us hold off on merging it in until we can discuss with the group. Nice fix!

src/js/add_input_icon.js
Vinnl: Yeah that's definitely something to consider, depending on whether that interferes with host styling on the wrapped input. Additionally, I think we'd want to consider using shadow DOM for at least the Relay icon, to avoid all the `!important`s in our stylesheet.

maxxcrawford: Ahhh. I love that it. I'll file that for a seperate enhancement. Great idea!

maxxcrawford:
@groovecoder @lesleyjanenorton I want to get y'alls take on switching out from a `div` to a `span`.

groovecoder:
Whew, I'm out of my depth on this one. I defer to y'all - @maxxcrawford @Vinnl @codemist @lesleyjanenorton .

groovecoder (approved):

Vinnl:
I'll just close this one - I don't expect it'll fix too many webcompat issues, and the burden of this lying around is bigger than the upside of the change itself :)

PR #151, opened 2021-10-05
DRAFT: Set/Sync Labels with Website Storage 

maxxcrawford:


maxxcrawford:
> Also a thought occurred to me - we won't do any data sync between server & add-on unless the user visits their profile page? 

Good catch. @Vinnl called this out too. I revised the panel UX to go to the settings page when they click "Allow" or just move on to the next onboarding panel if they click "Decline".

PR #195, opened 2021-10-26
Publish stage extension as artifact

Vinnl:
Although not an actual GitHub release (first steps towards that added in b130d7e), this at least makes the built stage extension available for download, so people don't have to run the steps themselves.

For an example, see https://github.com/mozilla/fx-private-relay-add-on/actions/runs/1386304980.

### DIFF
--- a/.github/workflows/sign-and-release-to-github.yml
+++ b/.github/workflows/sign-and-release-to-github.yml
@@ -52,11 +52,30 @@ jobs:
         id: release_notes
         run: echo ::set-output name=release_notes::$(git log --no-merges --pretty=format:"%h %s" ${{ steps.version.outputs.version }}^..${{ steps.version.outputs.version }})
 
-      - name: "Release"
-        uses: softprops/action-gh-release@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/sign-and-release-to-github.yml
```diff
@@ -52,11 +52,36 @@ jobs:
         id: release_notes
         run: echo ::set-output name=release_notes::$(git log --no-merges --pretty=format:"%h %s" ${{ steps.version.outputs.version }}^..${{ steps.version.outputs.version }})
 
-      - name: "Release"
```
groovecoder, line 55: Is this running the upload artifact as part of the release notes step?

groovecoder (approved):

.github/workflows/sign-and-release-to-github.yml
Vinnl: No, it's a new step (due to the `-`); it just doesn't have a name.

PR #199, opened 2021-11-02
Use a custom element instead of <div>s and <span>s

Vinnl:
Supersedes #144. This replaces all our instances of elements without semantic behaviour (i.e. `<span>`s and `<div>`s) with a behaviourless custom element. This might avoid overly generic CSS selectors in host web pages from matching on them, and thus might avoid some webcompat issues.

### DIFF
--- a/src/css/relay.css
+++ b/src/css/relay.css
@@ -340,6 +340,7 @@ button.fx-relay-menu-generate-alias-btn[disabled] {
 }
 
 .fx-relay-email-input-wrapper {
+  display: inline-block;
   position: relative;
   pointer-events: none;
   padding: 0 !important;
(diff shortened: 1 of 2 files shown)


groovecoder:
This seems like a solid approach, though I'd like to see us test thru some of the biggest sites on [our sheet of compat sites](https://docs.google.com/spreadsheets/d/1tKCK0CayEY2WJo5WLSxXhAazOCy9OoCCaPCanvlS2PE/edit#gid=0) before we merge this?

maxxcrawford (commented):
Is this still valid? LMK and I can review.

Vinnl:
@maxxcrawford Yeah, should still be valid I think - just rebased so there are no conflicts anymore.

Vinnl:
I'll just close this one - I don't expect it'll fix too many webcompat issues, and the burden of this lying around is bigger than the upside of the change itself :)

PR #201, opened 2021-11-05
Privacy notice update notification

Vinnl:
l10n dependency: https://github.com/mozilla-l10n/fx-private-relay-add-on-l10n/pull/13

This adds a notice to notify users about the upcoming privacy policy change that clarifies that their emails will be temporarily cached to avoid data loss.

### DIFF
--- a/src/js/popup.js
+++ b/src/js/popup.js
@@ -56,28 +56,28 @@ function premiumFeaturesAvailable(premiumEnabledString) {
   return false;
 }
 
-async function isServerStoragePromptPanelRelevant() {
-  const { serverStoragePrompt } = await browser.storage.local.get(
-    "serverStoragePrompt"
-  );
-
-
-  const serverStoragePref = await browser.runtime.sendMessage({
-    method: "getServerStoragePref"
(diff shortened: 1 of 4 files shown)


src/js/popup.js
```diff
@@ -56,31 +56,31 @@ function premiumFeaturesAvailable(premiumEnabledString) {
...
+    if (!serverStoragePref && !serverStoragePrompt) {
+      return true;
+    }
+
+    return false;
+  },
   hide: () => {
     const serverStoragePanelWrapper = document.querySelector(
-      ".server-storage-wrapper"
+      ".js-server-storage-wrapper"
```
maxxcrawford, line 83: Nice clean-up! 🧹

maxxcrawford (approved):
Clean PR. Nice fix on the `js-` classes. Good to go pending one question with Tony: 

**Is it expected to get double-prompted?** 

When testing on an account that had not yet opt in to label syncing, I first saw the privacy notice, dismissed it and THEN got the sync data question afterwards.

PR #212, opened 2021-11-18
Resize Popup Panel Images for Localized Strings

codemist:
This fixes https://github.com/mozilla/fx-private-relay/issues/1363 and resizes images depending on the length of the localized strings, in order to keep the 'Manage All Aliases' CTA always visible.


Vinnl (approved):
Ugh, pop-up sizing is a pain. Thanks for picking this up.

PR #214, opened 2021-11-19
Group JS files by use site

Vinnl:
This will hopefully make it easier to determine which file
communicates with which other file.

After this is merged I'll update existing PRs to reflect this change.


maxxcrawford (approved):
I have one request – Can we update `badge` to be more descriptive that it appears/runs on other websites? Something like `websites` or `other-websites` (whatever works for you). 

This is a great idea!

Vinnl:
Fixed by 60283f28baa419cb84f9a1a43b9457716643de0b.

PR #217, opened 2021-11-22
Don't ask to get Premium if that's impossible

Vinnl:
If Premium is not available in the user's country, this
makes it so they won't see suggestions to get Premium (which would
only end up in a bounce during the checkout process).

The API that lists which countries Premium is available in is only called once every seven days at most. This might mean that we only show the calls to upgrade to Premium a week after it's become available in an additional country, unless we update the extension beforehand. (/cc @groovecoder in case you've got opinions about server load.)

This depends on https://github.com/mozilla/fx-private-relay/pull/1374, so if you want to test this locally, make sure you've got that checked out for the server.

### DIFF
--- a/src/css/popup.css
+++ b/src/css/popup.css
@@ -288,7 +288,8 @@ main-panel {
 }
 
 .aliases-remaining {
-  margin: 0;
+  /* If the "Get unlimited aliases" prompt is hidden, this should be centered: */
+  margin: 0 auto;
   font-family: 'InterUI';
   background: white;
 }
--- a/src/js/background/background.js
+++ b/src/js/background/background.js
@@ -20,6 +20,39 @@ browser.runtime.onInstalled.addListener(async () => {
   }
 });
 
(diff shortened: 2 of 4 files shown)


maxxcrawford (commented):
I'm waiting on some testing steps for [this PR](https://github.com/mozilla/fx-private-relay/pull/1376#pullrequestreview-818290667) so that I can test this one accordingly.

maxxcrawford (approved):
I was still unable to get it organically to return false — however, I manually went to each query (in `popup/popup.js`, `other-websites/add_input_icon.js`, `background/background.js`) and each hid accordingly. 

As long as the query is sound, the correct things should hide. 

Note – This PR conflcits with https://github.com/mozilla/fx-private-relay-add-on/pull/221. I can handle the conflict!

Vinnl:
Hmm, that is weird, as only changing your Accept-Language should do it when running locally. But yeah, that logic is outside of this PR, so I'll merge i t.

PR #221, opened 2021-12-01
Feature: Re-use/autofill previously generated alias

maxxcrawford:
### Summary: 

This feature allows users to re-use aliases via context menu (alt-click)

### Major changes: 
- Moved all context menu code to separate background script 
- For non-premium accounts, the "Generate New Alias" will now stay visible but disabled after user has created maximum amount of aliases
- Added new top-level menu item: "Manage all aliases" that takes user to Relay account page 

#### TODO: 
- [x] Migrate strings to L10N 
- [x] Add logic for site-specific aliases
- [x] (Optional) Work with API to query aliases each interaction item rather than relay on local storage data
- [x] Add "Logged in" check during init
- [x] Only call `relayContextMenus.utils.getAliases` once during init. 
- [x] BUG: If menu is called twice, "Use on this site…" breaks. 

## Testing Steps:

- Start with a non-premium account with ZERO aliases created 
- Navigate to login form on a website (Ex: [Spotify](https://accounts.spotify.com/en/login), [Imgur](https://imgur.com/signin))
- Right/alt-click on Email field
- **Expected Menu:** Three items: Generate Alias, Mange All Aliases, Get Unlimited Aliases
- Right/alt-click: Generate alias
- **Expected:** New alias is created and populated into field. (You can validate this on the dashboard too) 
- Clear field, refresh page. 
- Right/alt-click on Email field again.
- **Expected menu:** should now include the Generate Alias, Use existing alias from this site…, Mange All Aliases, Get Unlimited Aliases
- Go to dashboard.
- Create a maximum amount of aliases, give some labels
- Go to a different site than the first 
- Right/alt-click on Email field
- **Expected menu** Generate Alias, Recent Alias, Mange All Aliases, Get Unlimited Aliases
  - Generate Alias should be greyed out
- Upgrade user account premium
- Right/alt-click on Email field
- **Expected menu** Generate Alias, Recent Alias, Mange All Aliases
  - Generate Alias should be back to normal

## Screenshots

_Note – The yellow background is OS-specific theme highlight and not part of this PR/code base_  

Non-premium Account: Generate New Alias is enabled
image

Non-premium Account: Generate New Alias is disabled as max alias count is reached 
image

Premium Account: Generate New Alias is enabled (and upgrade button is removed)
image

Premium Account: Reuse alias for site where no previous aliases were generated for: 
image

Premium Account: Reuse alias for site where some previous aliases were generated for: 
image


src/js/background/background.js
```diff
@@ -1,5 +1,3 @@
-(async function () {
```
maxxcrawford, line 1: Review note: As part of moving all the context menu code to its own file — wrapping the entire file in an async IFEE statement was no longer necessary.

src/js/background/background.js
```diff
@@ -166,149 +210,8 @@ async function makeRelayAddress(description = null) {
...
-  // If they revisit the settings page, it resets so that it only runs once again. 
-  if (!settingsRefresh) {
-    browser.storage.local.set({ settingsRefresh: true });
-
-    browser.tabs.query({ url: "http://127.0.0.1/*" }, function (tabs) {
-      for (let tab of tabs) {
-        browser.tabs.reload(tab.id);
-      }
-    });
-  }
```
maxxcrawford, line 311: Reviewer note: This code was moved to its own partial, and was then further refactored.

src/js/background/context-menu.js
```diff
@@ -0,0 +1,388 @@
...
+// TLDR: We make sure the menu is still open when performing the refresh command, as to not waste resoruces. 
+// This method is taken from https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus/onShown documentation. 
+let lastMenuInstanceId = 0;
+let nextMenuInstanceId = 1;
+
+// Events
+browser.menus.onShown.addListener(async (info, tab) => {
+  
+  if (!info.menuIds.includes("fx-private-relay-generate-alias") ) {
+    // No Relay menu items exist. Stop listening.
```
Vinnl, line 324: If you actually want to _stop_ listening, shouldn't this need a `.removeListener`?

src/js/background/context-menu.js
```diff
@@ -0,0 +1,388 @@
...
+  }
+
+  let menuInstanceId = nextMenuInstanceId++;
+  lastMenuInstanceId = menuInstanceId;
+
+  const domain = relayContextMenus.utils.getHostnameFromUrlConstructor(tab.url);
+  await relayContextMenus.init(domain);
+
+  if (menuInstanceId !== lastMenuInstanceId) {
+    return; // Menu was closed and shown again.
```
Vinnl, line 335: I don't get this - wouldn't it implicitly return after this `if` statement anyway? Why do you need this?

src/js/background/context-menu.js
```diff
@@ -0,0 +1,388 @@
...
+// Events
+browser.menus.onShown.addListener(async (info, tab) => {
+  
+  if (!info.menuIds.includes("fx-private-relay-generate-alias") ) {
+    // No Relay menu items exist. Stop listening.
+    return;
+  }
+
+  let menuInstanceId = nextMenuInstanceId++;
+  lastMenuInstanceId = menuInstanceId;
```
Vinnl, line 329: IMHO it's easier to follow if the reader doesn't need to know the evaluation order of the different operators :)

```suggestion
  let menuInstanceId = nextMenuInstanceId;
  lastMenuInstanceId = menuInstanceId;
  nextMenuInstanceId++;
```

src/js/background/context-menu.js
```diff
@@ -0,0 +1,388 @@
...
+
+  if (info.menuItemId.startsWith("fx-private-relay-use-existing-alias_")) {
+    await relayContextMenus.listeners.onFillInAddressWithAliasId(info, tab);
+  }
+
+});
+
+(async () => {
+  await relayContextMenus.init();
+})();
```
Vinnl, line 387: As in the other file, does this need the wrapping?

```suggestion
relayContextMenus.init();
```

src/js/background/context-menu.js
```diff
@@ -0,0 +1,388 @@
...
+      return websiteMatch
+    },
+    getHostnameFromRegex: (url) => {
+      // TODO: Unused function
+      // https://stackoverflow.com/a/54947757
+      // run against regex
+      const matches = url.match(/^https?\:\/\/([^\/?#]+)(?:[\/?#]|$)/i);
+      // extract hostname (will be null if no match is found)
+      return matches && matches[1];
+    },
```
Vinnl, line 243: If it's unused, might as well remove it? Feels more brittle than deferring to the browser for this anyway.

```suggestion
```

src/js/background/context-menu.js
```diff
@@ -0,0 +1,388 @@
...
+        
+        const filteredAliases = opts.currentWebsite
+          ? relayContextMenus.utils.getSiteSpecificAliases(
+              aliases,
+              opts.currentWebsite
+            )
+          : relayContextMenus.utils.getMostRecentAliases(aliases);
+
+        // // Cache server set locally
+        // browser.storage.local.set({ relayAddresses: aliases });
```
Vinnl, line 180:
```suggestion
```

src/js/background/context-menu.js
```diff
@@ -0,0 +1,388 @@
...
+
+    // Create "Manage all aliases" link
+    relayContextMenus.menus.create(staticMenuData.manageAliases);
+
+    // Generate upgrade menu item for non-premium users
+    const canUserUpgradeToPremium = await relayContextMenus.utils.getUserStatus.upgradeToPremium();
+    if (canUserUpgradeToPremium) relayContextMenus.menus.create([staticMenuData.upgradeToPremiumSeperator, staticMenuData.upgradeToPremium]);
+
+    // Set listerners
+    await browser.storage.onChanged.addListener(relayContextMenus.listeners.onLocalStorageChange);
```
Vinnl, line 87: Is `addListener` async?

Vinnl (commented):
I left some inline comments that are fairly straightforward, so feel free to incorporate them or not as you see fit. UX-wise, I think it might be good if the greyed-out "Generate new alias" would give some indication as to _why_ it's disabled - maybe something to bring up with Eduardo for a follow-up PR.

Unfortunately, I did find a couple of bugs, so I think those would have to be resolved before merging this:

If I disable my only alias, I do get the label "Use existing alias for this site", but it's empty:

image

The same happens if I then delete that alias. Also, I then get a "Recent aliases" item without a submenu:

image

And if I then click "Generate new alias", the new alias is listed under aliases for this site, but "Recent aliases" still doesn't have a submenu:

image

Hitting "Generate alias" multiple times in a row without refreshing lists those aliases under "Recent aliases", but not under "Recent for this site":

image

src/js/background/context-menu.js
maxxcrawford: `browser.menus.onShown` fires for all alt-clicks, so this filters it. This is a bit of pre-optimization, if we were to update the contexts in which Relay is shown (currently being set to the default, where it could be set to only editable fields, like inputs)

src/js/background/context-menu.js
maxxcrawford: I don't think this is actually required anymore — it came from this code example: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus/onShown

TODO: Remove all-of-this.

src/js/background/context-menu.js
maxxcrawford: ~Because it's an async function, and needs to be called in an [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) way. When `relayContextMenus.init();` runs (without the `await`), we get an error.~

Yeah, this works great! :)

maxxcrawford: TIL You don't always have to add `await` to `async` functions.

src/js/background/context-menu.js
maxxcrawford: ~the function it's calling is: `relayContextMenus.listeners.onLocalStorageChange`.~

Welp, my habit of `await`'ing all the things sure is starting to show.

maxxcrawford: ~the function it's calling is: `relayContextMenus.listeners.onLocalStorageChange`.~

Welp, my habit of `await`'ing all the things sure is starting to show.

src/js/background/context-menu.js
Vinnl: Haha :grin: The thing to keep in mind is that a function marked as `async` just returns a Promise behind the scenes, and `await`ing it is just a way to run whatever comes after in that Promise's `.then`.

src/js/background/context-menu.js
maxxcrawford: Removed in 787f56125a7bdf2aa0bd51eb60f64daecf080440

src/js/background/context-menu.js
maxxcrawford: Removed in 787f56125a7bdf2aa0bd51eb60f64daecf080440

src/js/background/context-menu.js
maxxcrawford: Removed in 787f56125a7bdf2aa0bd51eb60f64daecf080440

src/js/background/context-menu.js
maxxcrawford: Removed in 787f56125a7bdf2aa0bd51eb60f64daecf080440

src/js/background/context-menu.js
```diff
@@ -0,0 +1,388 @@
...
+      
+      function checkAvailability(aliases, domain) {
+        return aliases.some(function(alias) {
+            return domain === alias.generated_for;
+        });
+      }
+
+      const websiteMatch = checkAvailability(relayAddresses, website)
+
+      return websiteMatch
```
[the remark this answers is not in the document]
maxxcrawford: So clean. Revised in 787f56125a7bdf2aa0bd51eb60f64daecf080440

src/js/background/background.js
```diff
@@ -418,4 +292,6 @@ browser.runtime.onMessage.addListener(async (m) => {
   return response;
 });
 
+(async () => {
+  await displayBrowserActionBadge();
 })();
```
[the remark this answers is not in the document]
maxxcrawford: Revised in 787f56125a7bdf2aa0bd51eb60f64daecf080440

src/js/background/background.js
```diff
@@ -400,7 +273,8 @@ browser.runtime.onMessage.addListener(async (m) => {
       response = await sendMetricsEvent(m.eventData);
       break;
     case "rebuildContextMenuUpgrade":
-      await updateUpgradeContextMenuItem();
+      // console.log("rebuildContextMenuUpgrade-init");
```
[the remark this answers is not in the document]
maxxcrawford: Removed in 787f56125a7bdf2aa0bd51eb60f64daecf080440

src/js/background/context-menu.js
maxxcrawford: Removed in 787f56125a7bdf2aa0bd51eb60f64daecf080440

maxxcrawford: Context was from the code example here: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus/onShown

src/js/background/context-menu.js
```diff
@@ -0,0 +1,388 @@
...
+
+      const filteredAliases = array.filter(alias => alias.generated_for === domain);
+
+      // If 5 results for specific domain
+      if (filteredAliases.length > 5) { 
+        filteredAliases.length = 5;
+        return filteredAliases;
+      };
+
+      return filteredAliases
```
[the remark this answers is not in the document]
maxxcrawford: I love me some one-liner fixes. Revised in d9748b20854c90e83e4ac7f5884f6a6d2b87878b

src/js/background/context-menu.js
```diff
@@ -0,0 +1,388 @@
...
+      const { hostname } = new URL(url);
+      return hostname;
+    },  
+    getMostRecentAliases: (array)=> {
+      // Flipped to match the same order as the dashboard
+      array.reverse();
+
+      // Limit to 5
+      if (array.length > 5) { array.length = 5 };
+      return array;
```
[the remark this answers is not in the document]
maxxcrawford: Good catch. Revised in d9748b20854c90e83e4ac7f5884f6a6d2b87878b

src/js/background/context-menu.js
```diff
@@ -0,0 +1,388 @@
...
+    
+        if (numberOfAliasesCreated) {
+          aliasesRemaining = maxNumAliases - numberOfAliasesCreated.length;
+        }
+        
+        // The user cannot create and additional aliases.
+        if (aliasesRemaining < 1) return false;
+    
+        return true;
+      },
```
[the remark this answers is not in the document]
maxxcrawford: Yup. That's some legacy/refactor. Function name should be more readable. Updating function name. 

Revised in d9748b20854c90e83e4ac7f5884f6a6d2b87878b

src/js/background/context-menu.js
```diff
@@ -0,0 +1,388 @@
...
+      numberOfAliases: async () => {
+        const { relayAddresses } = await browser.storage.local.get("relayAddresses");
+        return relayAddresses.length;
+      },
+      upgradeToPremium: async()=> {
```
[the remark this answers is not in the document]
maxxcrawford: Agreed. Function name should be more readable. Revised in d9748b20854c90e83e4ac7f5884f6a6d2b87878b

I also updated the other function in the `getUserStatus` object: 
- `numberOfAliases` => `getNumberOfAliases`

src/js/background/context-menu.js
```diff
@@ -0,0 +1,388 @@
...
+    },
+    remove: async (id) => {
+      await browser.menus.remove(id);
+    }
+  }, 
+  utils: {
+    getAliases: async () => {
+
+      // TODO: This requires network access/server uptime. 
+      // Is there a way to test/fallback to local?
```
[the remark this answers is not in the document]
maxxcrawford: Refactored in 10ccb535e871e6e0ae12c1d6c98b8312b14d0abb

src/js/background/context-menu.js
maxxcrawford: Removed in 5cf283312e7756533cd32b3be3a0db633577a86d

src/js/background/background.js
```diff
@@ -20,6 +18,37 @@ browser.runtime.onInstalled.addListener(async () => {
...
+    method,
+    headers: headers,
+    body,
+  });
+
+  answer = await response.json();
+
+  // Cache alias list in local storage
+  // browser.storage.local.set({ relayAddresses: answer });
+
```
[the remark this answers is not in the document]
maxxcrawford: Removed in 5cf283312e7756533cd32b3be3a0db633577a86d

maxxcrawford:
@Vinnl I refactored the logic to create the "Use Existing Alias from this Site…" and "Recent Aliases"  parent menu items only where there are aliases to populate those menus. 

Hopefully this resolves your issues. There's some latency that exists when querying aliases from the server (~300ms), but that feels fast enough. LMK your thoughts!

src/js/background/context-menu.js
```diff
@@ -0,0 +1,371 @@
...
+      if (opts?.createExistingAliases) {
+        
+        const filteredAliases = opts.exisitingSite
+          ? relayContextMenus.utils.getSiteSpecificAliases(
+              aliases,
+              opts.currentWebsite
+            )
+          : relayContextMenus.utils.getMostRecentAliases(aliases, opts.currentWebsite);
+
+        // Only create the parent menu if we have will create sub-items
```
groovecoder, line 199:
```suggestion
        // Only create the parent menu if we will create sub-items
```

src/js/background/context-menu.js
```diff
@@ -0,0 +1,371 @@
...
+    await browser.menus.refresh();
+
+    return Promise.resolve(1)
+
+  },
+  listeners: {
+    onFillInAddressWithAliasId: async (info, tab) => {
+      // Trim the context menu id to get the alias reference ID
+      const selectedAliasId = info.menuItemId.replace(
+        "fx-private-relay-use-existing-alias_",
```
groovecoder, line 116: Feels like a magic string. Can we move it into a constant value to help keep us from accidentally changing it in only 1 place in the future?

src/js/background/context-menu.js
```diff
@@ -0,0 +1,371 @@
...
+          frameId: info.frameId,
+        }
+      );
+      
+    },
+    onLocalStorageChange: async (changes, area) => {
+      let changedItems = Object.keys(changes);
+      for (let item of changedItems) {
+        if (item === "relayAddresses") {
+          await relayContextMenus.init();
```
groovecoder, line 145: `init` calls `browser.storage.onChanged.addListener(relayContextMenus.listeners.
onLocalStorageChange);` again - is that safe?

src/js/background/context-menu.js
```diff
@@ -0,0 +1,371 @@
...
+  menus: {
+    create: async (data, opts=null, aliases) => {     
+      // If multiple items need to be created: 
+      if (Array.isArray(data)) {
+        data.forEach(async (menu) => {
+          await browser.menus.create(menu, relayContextMenus.utils.onCreatedCallback);
+        });
+        
+        return Promise.resolve(1)
+      }
```
groovecoder, line 187: I only see a single place where we call `relayContextMenus.menus.create` with an array argument. (i.e., `relayContextMenus.menus.create([staticMenuData.upgradeToPremiumSeperator, staticMenuData.upgradeToPremium])` ...

And this type-specific block completely changes the behavior of the function - e.g., if you pass an array argument, you can't make use of the `opts` or `aliases` argument values.

Can we either: 

1. remove this array-specific block and just call:
```js
relayContextMenus.menus.create(staticMenuData.upgradeToPremiumSeperator);
relayContextMenus.menus.create(staticMenuData.upgradeToPremium);
```

2. move this array-specific block to a new method:
```js
    createMultiple: async (data) => {     
      // If multiple items need to be created: 
      data.forEach(async (menu) => {
        await browser.menus.create(menu, relayContextMenus.utils.onCreatedCallback);
      });
      return Promise.resolve(1)
    }
```

src/js/background/context-menu.js
```diff
@@ -0,0 +1,371 @@
...
+        const { maxNumAliases } = await browser.storage.local.get("maxNumAliases");
+        const { relayAddresses } = await browser.storage.local.get("relayAddresses");
+
+        return (maxNumAliases - relayAddresses.length) > 0;
+      },
+      getNumberOfAliases: async () => {
+        const { relayAddresses } = await browser.storage.local.get("relayAddresses");
+        return relayAddresses.length;
+      },
+      canUpgradeToPremium: async()=> {
```
groovecoder, line 292: Should this try to account for what country the user is in? Or we can make that a follow-up later?

src/js/background/context-menu.js
```diff
@@ -0,0 +1,371 @@
...
+        const { premium } = await browser.storage.local.get("premium");
+      
+        // Note: If user is already premium, this will return false.
+        return !premium;
+      },
+    },
+    onCreatedCallback: ()=> {
+      // Catch errors when trying to create the same menu twice.
+      // The browser.menus API is limited. You cannot query if a menu item already exists.
+      // The error it throws does not show up to the user.
```
groovecoder, line 302: What error does it show? We can't `try/catch` it?

src/js/background/context-menu.js
```diff
@@ -0,0 +1,371 @@
...
+// Events
+browser.menus.onShown.addListener(async (info, tab) => {
+  
+  if (!info.menuIds.includes("fx-private-relay-generate-alias") ) {
+    // No Relay menu items exist. Stop listening.
+    return;
+  }
+
+  const domain = relayContextMenus.utils.getHostnameFromUrlConstructor(tab.url);
+   await relayContextMenus.init(domain);
```
groovecoder, line 320: white-space nit:
```suggestion
  await relayContextMenus.init(domain);
```

groovecoder (changes requested):
Some changes requested. Overall looks good. I didn't spot-check yet.

src/js/background/context-menu.js
Vinnl: That was dependent on #217 - resolved that by merging in those changes in 54c5092773dbac5896f992492f0d64621c04efba.

src/js/background/background.js
```diff
@@ -20,6 +18,34 @@ browser.runtime.onInstalled.addListener(async () => {
...
+    mode: "same-origin",
+    method,
+    headers: headers,
+    body,
+  });
+
+  answer = await response.json();
+  
+  return answer;
+}
```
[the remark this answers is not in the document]
Vinnl: Unfortunately I think that will be somewhat problematic, given that content scripts and background scripts run in a different context and neither supports module imports yet, so it's a bit hard to share code between them. Could be easier if we have a bundler setup that wires them together, but of course that comes with its own set of caveats.

src/js/background/context-menu.js
```diff
@@ -0,0 +1,371 @@
...
+      return;
+    }
+
+    // Reset any previously created menus
+    await browser.menus.removeAll();
+
+    // Generate aliases menu item
+    // If a user is maxed out/not premium, the generate item will be disabled.
+    const canUserGenerateAliases = await relayContextMenus.utils.getUserStatus.canGenerateMoreAliases();
+    canUserGenerateAliases ? relayContextMenus.menus.create(staticMenuData.generateAliasEnabled) : relayContextMenus.menus.create(staticMenuData.generateAliasDisabled);
```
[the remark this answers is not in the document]
Vinnl: 0180514

(I did it slightly differently to ensure go-to-definition for the menu items still works, hope that's OK.)

src/js/background/context-menu.js
Vinnl: 9a23436

src/js/background/context-menu.js
Vinnl: 31afe95

src/js/background/context-menu.js
Vinnl: 67d2c48

src/js/background/context-menu.js
Vinnl: fecbc65

Vinnl (commented):
I resolved conflicts arising from #217 in 54c5092773dbac5896f992492f0d64621c04efba (a merge commit, since I didn't want to break your local branch, but you can at least use it as inspiration when rebasing), and addressed Luke's comments insofar as I could (still a couple of open ones though).

Most of the bugs I mentioned earlier are resolved now. The following are still present, but their impact might be limited enough not to block it.

If I generate a new alias with a new user, the alias is listed under "Recent aliases" instead of "Recently used with this site" until I refresh:

image

Disabled aliases are still listed (though maybe that's intentional?):

https://user-images.githubusercontent.com/4251/145190949-fa197845-0ad1-407c-8571-fdd848c9aa1d.mp4

src/js/background/background.js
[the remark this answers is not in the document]
maxxcrawford: Agreed. Leaving this as is for right now.

src/js/background/context-menu.js
maxxcrawford: Note – Vincent did the right solve — just make'em two calls.

src/js/background/context-menu.js
maxxcrawford: ⚡

maxxcrawford:
> Disabled aliases are still listed (though maybe that's intentional?) 

@Vinnl Yes. I checked with Tony. He wants them listed for now. There's def. another round of UX that should go into this, but this should be good to land.

src/js/background/context-menu.js
maxxcrawford: This is basically a black-hole for errors when trying to make menu items. I pulled it out thinking it wasn't necessary, but was still getting some errors to show up in the console.

src/js/background/context-menu.js
maxxcrawford: There wasn't any visual side effects of calling this too many times but it's a good call out regardless. I conditionally remove it if it exists each time init is called. 

Revised in 2437ed69f721c92089de045f8b7c1dee8c5e1fff

groovecoder (approved):
Revisions look good. And agree with the reasons for non-revisions.

PR #226, opened 2021-12-06
Allow add-on detection without layout shifts

Vinnl:
Instead of having the website wait 500ms for the add-on to inject
data into it indicating that it is installed, this change will
allow it to simply add an `is-hidden-with-addon` class, and then
the add-on will inject CSS even before the JS is executed that will
ensure that that element is not visible (or `is-visible-with-addon`
for the inverse behaviour). This means we should no longer have
the page jump around when the "Install the add-on" banner is
displayed.

### DIFF
--- a/src/css/relay-website.css
+++ b/src/css/relay-website.css
@@ -1,0 +1,9 @@
+.is-visible-with-addon {
+    display: initial;
+    visibility: visible;
+}
+
+.is-hidden-with-addon {
+    display: none;
+    visibility: collapse;
+}
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -36,7 +36,13 @@
         "http://127.0.0.1/**"
       ],
       "js": ["js/relay.firefox.com/installation_indicator.js"]
-
+    },
+    {
+      "matches": [
+        "http://127.0.0.1/**"
+      ],
+      "css": ["css/relay-website.css"],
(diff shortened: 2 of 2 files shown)


maxxcrawford (changes requested):
What's the best way to test this?

Vinnl:
@maxxcrawford You can test it by also checking out https://github.com/mozilla/fx-private-relay/pull/1402, then making sure that the banner to install the add-on is not visible with the add-on installed, but _is_ visible if it's not (and you're in Firefox).

maxxcrawford (changes requested):
Thanks for the test steps — on Firefox without it loaded, everything works as expected.

I did notice this issue in Chrome. The current UX is that we recommend the user to try Firefox if they're visiting from a browser that doesn't support our add-on (read: Firefox only right now).

Can you make sure it's hidden entirely if not in Firefox? 

image

Vinnl:
@maxxcrawford Ooh, good catch, didn't think of that. I fixed that in https://github.com/mozilla/fx-private-relay-add-on/pull/226. Unfortunately that does mean there's still a layout shift when the banner _is_ present, but at least it'll be gone when it's not. (And it'll be entirely gone if we only have client-side rendering, since then the whole page will be rendered at the same time.)

Vinnl:
This was included in #249.

PR #242, opened 2021-12-10
Fix #239 - Only flip order of alias array if synced from the server

### LINKED ISSUE #239: The order of aliases in the context submenu is not consistent
#### Build:
- Stage;

#### Affected Platforms:
- Windows 10;

#### Browsers tested on:
- Firefox, latest version;

#### Prerequisites:
- Be signed in into a Firefox Relay account;
- Do not have any aliases generated;

#### Steps to reproduce:
1. Go to the Relay dashboard;
2. Generate 3 aliases and do not touch them;
3. Observe the order of the aliases;
4. Go to any site that displays a fill-able email field (e.g. Amazon registration);
5. Right click on the field, hover on the Firefox Relay submenu, then hover on the Recent Aliases submenu;
6. Observe the order the aliases in the submenu;
7. Return to the dashboard and compare the orders displayed in the submenus with the one on the dashboard;

#### Expected result:
- The alias orders in the submenus is the same as the order on the dashboard;

#### Actual result:
- The alias orders in the submenus is the opposite of the order on the dashboard;

#### Notes:
- This happens for both the “Recent Aliases” list and the “Existing Alias” list;
- Usually the list is displayed in the opposite order of the dashboard but sometimes, on some sites, the list is displayed correctly. But eventually, as time passes, all sites seem to update and display the opposite order. Not constant, but Facebook seemed to return the correct oreder most of the times;
- See attached media:
Initial alias order, Dashboard
alord1 1
Alias order submenu, Facebook
alord1 2
Alias order submenu, Amazon
alord1 3
Alias order submenu, Google accounts
alord1 4

maxxcrawford:
This fixes #239 

STR: 

- Create multiple aliases in the dashboard. Name them 1-2-3 from top to bottom. 
- Go to any site, right click on email form
- **Expected:** The order should be 1,2,3 from top to bottom. 
- Go to dashboard settings > uncheck data opt-in. 
- Go back to dashboard. Confirm aliases are still named 1,2,3 (from top to bottom) 
- Go to any site, right click on email form
- **Expected:** The order should be 1,2,3 from top to bottom.

### DIFF
--- a/src/js/background/context-menu.js
+++ b/src/js/background/context-menu.js
@@ -217,12 +217,19 @@ const relayContextMenus = {
       // Loop Through Existing Aliases
       if (opts?.createExistingAliases) {
         
+        const shouldAliasOrderBeReversed = await getServerStoragePref();
+
         const filteredAliases = opts.exisitingSite
           ? relayContextMenus.utils.getSiteSpecificAliases(
               aliases,
-              opts.currentWebsite
+              opts.currentWebsite,
+              shouldAliasOrderBeReversed
             )
-          : relayContextMenus.utils.getMostRecentAliases(aliases, opts.currentWebsite);
+          : relayContextMenus.utils.getMostRecentAliases(
+              aliases,
+              opts.currentWebsite,
+              shouldAliasOrderBeReversed
+            );
 
         // Only create the parent menu if we will create sub-items
         if (filteredAliases.length > 0) {
@@ -281,9 +288,12 @@ const relayContextMenus = {
       const { hostname } = new URL(url);
       return hostname;
     },  
-    getMostRecentAliases: (array, domain)=> {
-      // Flipped to match the same order as the dashboard
-      array.reverse();
+    getMostRecentAliases: (array, domain, shouldFlipArray)=> {
+      // Flipped to match the same order as the dashboard if synced from the server
+      if (shouldFlipArray) {
+        array.reverse();
+      }
+      
 
       // Remove any sites that match the current site (inverse of getSiteSpecificAliases())
       const filteredAliases = array.filter(alias => alias.generated_for !== domain);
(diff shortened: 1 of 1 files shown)


src/js/background/context-menu.js
```diff
@@ -217,12 +217,19 @@ const relayContextMenus = {
       // Loop Through Existing Aliases
       if (opts?.createExistingAliases) {
         
+        const shouldAliasOrderBeReversed = await getServerStoragePref();
```
Vinnl, line 220: Could you add a comment here about _why_ the order should be revered if the labels are stored on the server? That's not clear to me :/

src/js/background/context-menu.js
```diff
@@ -281,20 +288,25 @@ const relayContextMenus = {
       const { hostname } = new URL(url);
       return hostname;
     },  
-    getMostRecentAliases: (array, domain)=> {
-      // Flipped to match the same order as the dashboard
-      array.reverse();
+    getMostRecentAliases: (array, domain, shouldFlipArray)=> {
+      // Flipped to match the same order as the dashboard if synced from the server
+      if (shouldFlipArray) {
```
Vinnl, line 293: I generally prefer turning optional parameters into named parameters. That way, if we have to add more parameteres in the future, you don't have to pass in a bunch of `undefined`s.

I also prefer using named parameters for boolean arguments (which often are optional), just because it's clearer to people reading code at the call site to determine what that parameter does.

```suggestion
    getMostRecentAliases: (array, domain, options = {})=> {
      // Flipped to match the same order as the dashboard if synced from the server
      if (options.shouldFlipArray) {
```

So this would be called as:

```js
getMostRecentAliases(aliases, currentWebsite, { shouldAliasOrderBeReversed: true })`.
```

Likewise for the other instance of this parameter.

Vinnl (approved):
Two suggestions for you to consider, but it does what it needs to do :)

src/js/background/context-menu.js
maxxcrawford: I dig it! I was already using an options argument on the parent function. Good call out on using it in all the spots!

src/js/background/context-menu.js
maxxcrawford: Added in 54a57e0b85ff4b99e59537b9d7a319f25cb07bb2

src/js/background/context-menu.js
maxxcrawford: Revised in 54a57e0b85ff4b99e59537b9d7a319f25cb07bb2

maxxcrawford:
@Vinnl Thanks for the thoughtful review! 🏁

PR #246, opened 2021-12-15
Remove the `premiumEnabled` feature flag

Vinnl:
Premium has now been enabled for everyone. This is a companion PR to https://github.com/mozilla/fx-private-relay/pull/1428,  which should be merged after this is released.


maxxcrawford (changes requested):
I don't see exactly what's breaking the logic here, but if I this branch on the add-on and https://github.com/mozilla/fx-private-relay/pull/1428 locally, I do _not_ get the upgrade prompts. Can you verify this is in working order?

PR #248, opened 2022-01-04
Link "Get unlimited aliases" to interstitial

Vinnl:
Instead of jumping directly into the purchasing flow, this first
brings people to a page that explains what Premium is and what it
will get them.


maxxcrawford (approved):
This looks and works great! We should annotate this change in Google Analytics when it deploys (since it'll move our metrics some)

codemist (approved):
Looks good!

Vinnl:
@maxxcrawford Can you track that? (Possibly pairing with me.) I don't know what that is, and don't have access to GA.

PR #252, opened 2022-01-12
Create a deployment for every add-on push

Vinnl:
This creates a GitHub Deployment whenever we push. If you then create a pull request, you can see it listed under "Environments", as below. Unfortunately, it doesn't seem to be possible to link to the artifact directly, so the "View deployment" link currently links to the Workflow run overview, which lists the add-on.

What it looks like when it's in progress:

image

And when built:

image

The "deployed" link is to the artifact:

image

I think if additional deployments are done after the PR is created, they are listed sa events below (in a similar style to "self-assigned this 3 minutes ago").

### DIFF
--- a/.github/workflows/sign-and-release-to-github.yml
+++ b/.github/workflows/sign-and-release-to-github.yml
@@ -2,15 +2,53 @@ name: continuous prerelease to GitHub
 
 on:
   push:
-    branches: [ main ]
-
+    branches: [ "*" ]
   workflow_dispatch:
 
 jobs:
+  prepare-deployment:
+    runs-on: ubuntu-latest
+    outputs:
+      deployment-id: ${{ fromJson(steps.create-deployment.outputs.result).data.id }}
+    steps:
+      - name: Create GitHub Deployment
+        id: create-deployment
+        uses: actions/github-script@v5
+        with:
+          script: |
+            return await github.rest.repos.createDeployment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              ref: context.sha,
+              environment: "review",
+              transient_environment: true,
+              auto_merge: false,
+              // The deployment runs in parallel with CI, so status checks will never have succeeded yet:
+              required_contexts: [],
+            });
+
+
   sign-tag-release:
     runs-on: ubuntu-latest
-
+    needs: [prepare-deployment]
     steps:
+      - name: Mark GitHub Deployment as in progress
+        id: start-deployment
+        uses: actions/github-script@v5
+        env:
+          DEPLOYMENT_ID: "${{ needs.prepare-deployment.outputs.deployment-id }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/sign-and-release-to-github.yml
```diff
@@ -2,15 +2,56 @@ name: continuous prerelease to GitHub
...
-
+    branches: [ "*" ]
   workflow_dispatch:
 
 jobs:
-  sign-tag-release:
+  prepare-deployment:
     runs-on: ubuntu-latest
+    outputs:
+      tag-slug: ${{ steps.determine-npm-tag.outputs.tag-slug }}
```
groovecoder, line 12: Where does this come from? And it looks like we're not using it anywhere?

.github/workflows/sign-and-release-to-github.yml
```diff
@@ -2,15 +2,56 @@ name: continuous prerelease to GitHub
...
+          script: |
+            return await github.rest.repos.createDeploymentStatus({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              deployment_id: process.env.DEPLOYMENT_ID,
+              description: "Building add-on",
+              environment: "review",
+              log_url: `https://github.com/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`,
+              state: "in_progress",
+              mediaType: {previews: ["flash", "ant-man"]},
```
groovecoder, line 52: What are these media types? I didn't find them anywhere in the docs.

groovecoder (approved):
Just questions about how it's working, but the code looks good (nice find of `github-script` action!) and clearly the PR is deploying itself successfully.

.github/workflows/sign-and-release-to-github.yml
Vinnl: Whoops! I based this on [the workflows I setup at my previous place](https://github.com/inrupt/solid-client-js/actions/workflows/cd.yml), and this is a remnant of that. Removed in 35bdb1a.

.github/workflows/sign-and-release-to-github.yml
Vinnl: I think they were listed in the docs back when I first implemented this; they're used for API previews. Specifically, they're for fairly minor features we're using: [automatically labelling a Deployment as inactive](https://developer.github.com/changes/2016-04-06-deployment-and-deployment-status-enhancements/#automatic-creation-of-inactive-statuses) and [being able to list a Deployment as "in progress"](https://developer.github.com/changes/2018-10-16-deployments-environments-states-and-auto-inactive-updates/). I can't find any posts about those API's having graduated, but it also seems that the deployment still works just fine (including marking as "in progress") without them (d57000d), and they're no longer listed in the documentation examples, so I'll just keep them out.

PR #254, opened 2022-01-17
Prepare add-on for client-side rendering

Vinnl:
This adds two minor changes to the add-on to make it compatible with the React prototype I wrote:

1. It wraps the code in `get_profile_data.js` in a function called `run`. If the element `#profile-main` exists at the time the script is executed (this is the current state: the HTML is generated by Django before being sent to the browser), that function is immediately executed, so the behaviour is exactly the same as now. If that element is _not_ present, however (this is what happens when the React code is sent to the browser, which will then execute it to generate the DOM _after_ `get_profile_data.js` is run), then a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) will be added, which watches whether `#profile-main` is added to the DOM, and only _then_ executes `run`.
2. In `installation_indicator.js`, it loads the locally stored alias labels (if any), and injects them into the website by setting the `data-local-labels` attribute on the `<firefox-private-relay-addon>` element. The React version of the website I wrote handles local storage of labels by itself, and this allows it to import any existing local labels.

How to test:
- The add-on behaviour should not change when used with the website as it exists today.
- The add-on should also continue working at it was when switching to the `react-experiment` branch of the main website.

### DIFF
--- a/src/js/relay.firefox.com/get_profile_data.js
+++ b/src/js/relay.firefox.com/get_profile_data.js
@@ -1,419 +1,444 @@
 (async function () {
-  // Get the api token from the account profile page
-  const profileMainElement = document.querySelector("#profile-main");
-  const apiToken = profileMainElement.dataset.apiToken;
-  browser.storage.local.set({ apiToken });
+  const dahsboardInitializationObserver = new MutationObserver((mutations) => {
+    mutations.forEach(mutation => {
+      if (mutation.type !== "childList") {
+        return;
+      }
+      mutation.addedNodes.forEach(node => {
+        if (node.id !== "profile-main") {
+          return;
+        }
 
-  // API URL is ${RELAY_SITE_ORIGIN}/api/v1/
-  const { relayApiSource } = await browser.storage.local.get("relayApiSource");
+        // Once an element `#profile-main` is added, the dashboard is initialized
+        // with the data needed by the add-on; stop watching for futher changes,
+        // and run once:
+        dahsboardInitializationObserver.disconnect();
+        run();
+      });
+    });
+  });
+  if (document.getElementById("profile-main") === null) {
+    dahsboardInitializationObserver.observe(document.body, { childList: true, subtree: true });
+  } else {
+    run();
+  }
 
-  const apiProfileURL = `${relayApiSource}/profiles/`;
-  const apiRelayAddressesURL = `${relayApiSource}/relayaddresses/`;
+  async function run() {
+    // Get the api token from the account profile page
+    const profileMainElement = document.querySelector("#profile-main");
+    const apiToken = profileMainElement.dataset.apiToken;
+    browser.storage.local.set({ apiToken });
 
-  async function apiRequest(url, method = "GET", body = null, opts=null) {
+    // API URL is ${RELAY_SITE_ORIGIN}/api/v1/
+    const { relayApiSource } = await browser.storage.local.get("relayApiSource");
 
-    const cookieString =
-      typeof document.cookie === "string" ? document.cookie : "";
-    const cookieStringArray = cookieString
-      .split(";")
-      .map((individualCookieString) => individualCookieString.split("="))
-      .map(([cookieKey, cookieValue]) => [
-        cookieKey.trim(),
-        cookieValue.trim(),
-      ]);
+    const apiProfileURL = `${relayApiSource}/profiles/`;
+    const apiRelayAddressesURL = `${relayApiSource}/relayaddresses/`;
 
-    const [_csrfCookieKey, csrfCookieValue] = cookieStringArray.find(
-      ([cookieKey, _cookieValue]) => cookieKey === "csrftoken"
-    );
+    async function apiRequest(url, method = "GET", body = null, opts=null) {
 
-    browser.storage.local.set({ csrfCookieValue: csrfCookieValue });
+      const cookieString =
+        typeof document.cookie === "string" ? document.cookie : "";
+      const cookieStringArray = cookieString
+        .split(";")
+        .map((individualCookieString) => individualCookieString.split("="))
+        .map(([cookieKey, cookieValue]) => [
+          cookieKey.trim(),
+          cookieValue.trim(),
+        ]);
+
+      const [_csrfCookieKey, csrfCookieValue] = cookieStringArray.find(
+        ([cookieKey, _cookieValue]) => cookieKey === "csrftoken"
+      );
+
+      browser.storage.local.set({ csrfCookieValue: csrfCookieValue });
 
 
-    const headers = new Headers();
+      const headers = new Headers();
 
-    
-    headers.set("X-CSRFToken", csrfCookieValue);
-    headers.set("Content-Type", "application/json");
-    headers.set("Accept", "application/json");
-    
-    if (opts && opts.auth) {
-      const apiToken = await browser.storage.local.get("apiToken");
-      headers.set("Authorization", `Token ${apiToken.apiToken}`);
+
+      headers.set("X-CSRFToken", csrfCookieValue);
+      headers.set("Content-Type", "application/json");
+      headers.set("Accept", "application/json");
+
+      if (opts && opts.auth) {
+        const apiToken = await browser.storage.local.get("apiToken");
+        headers.set("Authorization", `Token ${apiToken.apiToken}`);
+      }
+
+
+      const response = await fetch(url, {
+        mode: "same-origin",
+        method,
+        headers: headers,
+        body,
+      });
+
+      answer = await response.json();
+      return answer;
     }
 
+    const serverProfileData = await apiRequest(apiProfileURL);
 
-    const response = await fetch(url, {
-      mode: "same-origin",
-      method,
-      headers: headers,
-      body,
+    browser.storage.local.set({
+      profileID: parseInt(serverProfileData[0].id, 10),
+      settings: {
+        server_storage: serverProfileData[0].server_storage,
+      },
     });
 
-    answer = await response.json();
-    return answer;
-  }
-
-  const serverProfileData = await apiRequest(apiProfileURL);
-
-  browser.storage.local.set({
-    profileID: parseInt(serverProfileData[0].id, 10),
-    settings: {
-      server_storage: serverProfileData[0].server_storage,
-    },
-  });
-
-  // Get the relay address objects from the addon storage
-  const addonStorageRelayAddresses = await browser.storage.local.get(
-    "relayAddresses"
-  );
-
-  const siteStorageEnabled = serverProfileData[0].server_storage;
-
-  const addonRelayAddresses =
-    Object.keys(addonStorageRelayAddresses).length === 0
-      ? { relayAddresses: [] }
-      : addonStorageRelayAddresses;
-
-  // Check if user is premium
-  const isPremiumUser = document
-    .querySelector("body")
-    .classList.contains("is-premium");
-  browser.storage.local.set({ premium: isPremiumUser });
-
-  // Loop over the addresses on the page
-  const dashboardRelayAliasCards = document.querySelectorAll(
-    "[data-relay-address]"
-  );
-
-  const localStorageRelayAddresses = [];
-
-  // Get FXA Stuff
-  const fxaSubscriptionsUrl = document.querySelector(
-    "firefox-private-relay-addon-data"
-  ).dataset.fxaSubscriptionsUrl;
-  const premiumProdId = document.querySelector(
-    "firefox-private-relay-addon-data"
-  ).dataset.premiumProdId;
-  const premiumPriceId = document.querySelector(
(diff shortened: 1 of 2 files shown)


src/js/relay.firefox.com/get_profile_data.js
```diff
@@ -1,419 +1,453 @@
...
+        dahsboardInitializationObserver.disconnect();
+        run();
+      });
+    });
+  });
+  if (document.getElementById("profile-main") === null) {
+    dahsboardInitializationObserver.observe(document.body, { childList: true, subtree: true });
+  } else {
+    run();
+  }
```
Vinnl, line 24: This diff looks intimidating, but the above is the only code that's new. Everything below is exactly the same as it was, except for being wrapped inside the `run()` function.

src/js/relay.firefox.com/installation_indicator.js
```diff
@@ -4,8 +4,27 @@
...
+        .filter(address => address.description.length > 0)
+        .map(address => ({
+          type: "random",
+          id: Number.parseInt(address.id, 10),
+          description: address.description,
+          address: address.address,
+        })
+      );
+      el.dataset.localLabels = JSON.stringify(localLabels);
+    }
```
Vinnl, line 27: This is the second change referred to in the pull request description. After migrating to the React implementation of the website, that will allow us to get rid of all the label code inside the add-on (i.e. `settings_refresh.js` and a lot of code in `get_profile_data.js`).

src/js/relay.firefox.com/get_profile_data.js
```diff
@@ -1,419 +1,453 @@
...
+
+      if (localStorageRelayAddresses.length === 0) {
+        // If we weren't able to scrape alias data from the page, that means
+        // the React version of the website (with a different DOM structure)
+        // has been deployed. Since the React version handles local storage of
+        // label data by itself, we don't need to add our own listeners,
+        // and we can just fetch label data from the API.
+        const serverRelayAddresses = await apiRequest(apiRelayAddressesURL);
+        localStorageRelayAddresses.push(...serverRelayAddresses);
+      }
```
Vinnl, line 445: This part is also new, and is part of the second change (54e5c27666a6021ff86d08fc3954179b4fc23802): whereas we scrape the DOM to get aliases and their labels with the old website, we just fetch them from the API in the React-based version.

src/js/relay.firefox.com/installation_indicator.js
```diff
@@ -4,8 +4,27 @@
 
 (async () => {
   localStorage.setItem("fxRelayAddonInstalled", "true");
-  document.querySelectorAll("firefox-private-relay-addon").forEach((el) => {
+  document.querySelectorAll("firefox-private-relay-addon").forEach(async (el) => {
     el.dataset.addonInstalled = "true";
+
+    // In the server-rendered version of the website, the add-on would store alias labels
+    // locally if server-side storage was disabled.
+    // In the React version of the website, the website handles local storage itself.
```
groovecoder, line 12: I don't follow this ... regardless of whether the website is React or not, we still want the add-on to store labels locally if server-side storage is disabled.

groovecoder (changes requested):
A few Q's before I get into spot-checking.

src/js/relay.firefox.com/get_profile_data.js
```diff
@@ -1,419 +1,453 @@
...
+        // Once an element `#profile-main` is added, the dashboard is initialized
+        // with the data needed by the add-on; stop watching for futher changes,
+        // and run once:
+        dahsboardInitializationObserver.disconnect();
+        run();
+      });
+    });
+  });
+  if (document.getElementById("profile-main") === null) {
+    dahsboardInitializationObserver.observe(document.body, { childList: true, subtree: true });
```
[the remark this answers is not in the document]
Vinnl: Yeah we could do that, but that would require us to first do a production deploy of the current website before distributing the updated add-on to the users.

It's also going to be very temporary, as this is currently only used to get the API token ([here](https://github.com/mozilla/fx-private-relay-add-on/blob/e099130eaf5cc8ce66c74c6cb127d008c4ba2b23/src/js/relay.firefox.com/get_profile_data.js#L28-L29)) - which we're moving away from with the new auth mechanism :) Once we've transitioned to that, we can go back to just executing `run()` immediately. So it seems like an acceptable trade-off to me, but happy to discuss.

src/js/relay.firefox.com/get_profile_data.js
```diff
@@ -1,419 +1,453 @@
...
+        };
 
-      localStorageRelayAddresses.push(relayAddress);
+        localStorageRelayAddresses.push(relayAddress);
+      }
+
+      if (localStorageRelayAddresses.length === 0) {
+        // If we weren't able to scrape alias data from the page, that means
+        // the React version of the website (with a different DOM structure)
+        // has been deployed. Since the React version handles local storage of
```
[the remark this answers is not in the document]
Vinnl: Good question. So `localStorageRelayAddress` gets populated in this for loop:

https://github.com/mozilla/fx-private-relay-add-on/blob/e099130eaf5cc8ce66c74c6cb127d008c4ba2b23/src/js/relay.firefox.com/get_profile_data.js#L228

That's looping over all elements matching `[data-relay-address]`:

https://github.com/mozilla/fx-private-relay-add-on/blob/e099130eaf5cc8ce66c74c6cb127d008c4ba2b23/src/js/relay.firefox.com/get_profile_data.js#L108-L111

That will work in the current site. In the new site, however, there are no elements matching that query selector, so the loop will never run. (This is fine, because it does just two things: fetch alias data, and set up local label storage. As I'll mention in the other comment, the add-on will no longer be responsible for local label storage. And this `if` block takes over the fetching of alias data.)

It does mean that in the current site, if the user has 0 aliases, the add-on will also do an API request to again fetch those zero aliases. That seems acceptable in the transitory period, but open to reconsider if you disagree.

src/js/relay.firefox.com/installation_indicator.js
Vinnl: I went with a slightly different approach in the React website (already discussed with @maxxcrawford): now, the website instead of the add-on stores the labels locally (i.e. in localStorage). It only does this if the add-on is installed and server-side storage is enabled.

The main advantage of this is that we have a single code path for editing labels, which just at the very end determines whether to store them locally or on the server. This avoids the situation we have now, where the add-on just reaches into the website DOM and manipulates it, which:

- means that changing the website can break the add-on's interaction with it
- results in different label behaviour with local vs server-side storage (e.g. there's some validation that's only applied in local storage today)
- requires add-on deployments to make changes

In this new approach, the add-on injects the existing labels, which the website then imports once, and from then on it handles local storage of labels.

src/js/relay.firefox.com/get_profile_data.js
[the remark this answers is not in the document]
groovecoder: Website prod deploys tend to happen sooner and more often than add-ons already, so I feel fine about that. And I'm willing to be the new auth mechanism will take a while longer than "very" temporary.

src/js/relay.firefox.com/installation_indicator.js
groovecoder: Oh, interesting. Sorry to throw a wrench in that plan, but what happens when someone with Relay extension or add-on uses "Clear site data" in their browser? That will remove all their alias labels? Even though they still have the Relay extension installed?

src/js/relay.firefox.com/get_profile_data.js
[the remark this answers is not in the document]
Vinnl: Hmm OK, so thinking about this some more, I don't even think we'll need an update to the current website, since the MutationObserver is only initialised for the new one. What I could do is update the React front-end to set a `data` attribute on the body the moment that `profile-main` is added to the DOM, so that the MutationObserver only needs to look for changes to that specific attribute. What do you think of that approach?

src/js/relay.firefox.com/installation_indicator.js
Vinnl: Yes, that would also remove the labels. That said, that would match expected behaviour for me, just like how I'm logged out of the Relay website when I do that.

In fact, it would even make sense to me to take this one step further: we can locally store alias labels even _without_ the add-on installed. The way I've currently implemented this, is to have an explicit check for if the add-on is installed, and if not, to just not display labels at all, so this could be achieved by simply removing that check.

src/js/relay.firefox.com/get_profile_data.js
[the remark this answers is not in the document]
maxxcrawford: > Yeah we could do that, but that would require us to first do a production deploy of the current website before distributing the updated add-on to the users.
> 
> It's also going to be very temporary, as this is currently only used to get the API token ([here](https://github.com/mozilla/fx-private-relay-add-on/blob/e099130eaf5cc8ce66c74c6cb127d008c4ba2b23/src/js/relay.firefox.com/get_profile_data.js#L28-L29)) - which we're moving away from with the new auth mechanism :) Once we've transitioned to that, we can go back to just executing `run()` immediately. So it seems like an acceptable trade-off to me, but happy to discuss.

This may no longer apply but we definitely can land a small change to the main website first. Is this still applicable?

src/js/relay.firefox.com/installation_indicator.js
maxxcrawford: That's interesting to do this without the add-on all together. 

Question: If a user clears site data right now, does that clear the add-on stored labels? If that's the case, we would need to either: 
- Provide a way to back up/export/import labels locally for private-conscious users
- Provide notice of this change before we actually implement it.

maxxcrawford (commented):
I commented on some open questions from @groovecoder. I think if we can land something (and deploy it) on the main website first, lets do that.

src/js/relay.firefox.com/get_profile_data.js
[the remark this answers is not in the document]
Vinnl: Nope!

> I don't even think we'll need an update to the current website

So the question is: do we want to explicitly add an attribute to the `<body>` at the same time we add `#profile-main` to the DOM? It's slightly more error-prone (in case "at the same time" happens to not be at the same time), but should be more performant.

src/js/relay.firefox.com/installation_indicator.js
Vinnl: I just cleared all history:

image

That does not clear the stored labels. I would think the group of people who

- Have disabled server-side label storage, and
- Have locally-stored labels, and
- Regularly clear site data, and
- Expect that to not remove their locally stored labels, which they primarily see in the website, and
- Would go through the hassle of exporting and importing their labels every time they do that

is in the low double digits, if not single digits (or even 0!), so building a custom export/import system seems like not worth the cost.

Adding a notice is less effort, but even if we only show that to people who have locally stored labels, I'd be a bit worried that the vast majority of them would never experience any issue with this, and thus such a warning might only serve to confuse them. But that's something we might want to bring up with Tony and Chris.

Vinnl: OK, so I just spent some time moving label storage back into the add-on. In addition to the concern above, a more pressing issue was that the add-on cannot use labels in e.g. context menus if they are stored in the website context.

Thus, I added 1df7cdd3. This works in cooperation with https://github.com/mozilla/fx-private-relay/pull/1518/commits/ab545456a1b8571e43ccde4253a646558de35125 and https://github.com/mozilla/fx-private-relay/pull/1518/commits/1df7cdd3de3a5b795cdee684940e7b1f32528622: when a label is updated on the website, it now triggers an event on the `<firefox-private-relay-addon>` element. The add-on listens for those events and updates its local cache of aliases. This preserves the fact that the add-on is no longer reaching into the website DOM, resulting in inconsistent flows, and sets a clear API boundary (the custom element) for the website and add-on to communicate through, while moving the label and `generated_for` data into the add-on.

maxxcrawford: Got'cha. I appreciate the pivot!

src/js/relay.firefox.com/get_profile_data.js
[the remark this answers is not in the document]
groovecoder: So the failure states are degraded performance vs. broken functionality? Let's play it safe first - i.e., add the mutation observers to `document.body` for now. If it causes significant performance problems in production, we know we can try this technique of adding the `data` attribute to `profile-main`.

src/js/relay.firefox.com/installation_indicator.js
groovecoder: Neat trick! @Vinnl and I discussed a bit today; this implementation seems safe enough - @Vinnl pointed out that this content script only runs on our own profile page.

groovecoder (changes requested):

groovecoder (approved):

PR #257, opened 2022-01-19
Fix MPP-1568: All upgrade links now point to the /premium interstitial page 

maxxcrawford:
This PR updates the "upgrade" links across the add-on. 

Testing: 
- Use account that is _not_ premium.
- Navigate to site with login.
- Alt/right click to display context menu
- Navigate to Firefox Relay section, click "Get Unlimited Aliases"
- **Expected:** It should open a new tab, pointed to /premium page  
- Click on Relay icon in email field
- Click on "Get Unlimited Aliases"
- **Expected:** It should open a new tab, pointed to /premium page

### DIFF
--- a/src/js/background/context-menu.js
+++ b/src/js/background/context-menu.js
@@ -374,13 +374,8 @@ browser.contextMenus.onClicked.addListener(async (info, tab) => {
         action: "click",
         label: "context-menu-get-unlimited-aliases",
       });
-      const { fxaSubscriptionsUrl, premiumProdId, premiumPriceId } =
-        await browser.storage.local.get([
-          "fxaSubscriptionsUrl",
-          "premiumProdId",
-          "premiumPriceId",
-        ]);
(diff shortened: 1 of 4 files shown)


Vinnl (approved):
Works on my machine, and code looks good. Only thing we could consider is getting rid of `fxaSubscriptionsUrl`, `premiumProdId` and `premiumPriceId` everywhere - there's at least unused references to them in `popup.js` (they already were unused before this PR). We might also consider removing them  from `get_profile_data` (as well as exposing them from the website).

maxxcrawford:
@Vinnl Good call-out. 

I know @tcinotto wanted/is-planning on doing some A/B testing on which link performs better (directly into subplat vs /premium interstitial page). We may need to leave it as is (in the API) for now. I'll go remove the unused get requests.

maxxcrawford:
Yeah – I'll leave that reference as is for now. 😄

PR #258, opened 2022-01-19
Fix #236 / MPP-1481: Show Correct Panel based on user's premium state

### LINKED ISSUE #236: Free user addon menu displayed for premium user
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit

**Browsers tested on:**
Firefox

**Preconditions:**
The user has a premium subscription.
Latest version of Relay addon is installed.

**Steps to reproduce:**

1. Navigate to Stage server and log in to Firefox Relay. 
2. Click the Firefox Relay addon icon in the toolbar.
4. Observe the menu.

**Actual result:**
The menu specific for free user is displayed instead of the one for premium user.
 
**Expected result:** 
The premium user addon menu is displayed.

free user menu displayed instead of premium

maxxcrawford:
Fixes #236 and [MPP-1481](https://mozilla-hub.atlassian.net/browse/MPP-1481)

Summary: There was an issue where premium users who opened the panel would be presented with  wrong panel/content.

Fix: Removed `premiumCountryAvailability` check to show premium panel. If the user has already upgraded, we can infer their country/state is not an issue

## Testing Steps
- Have a premium account already created
- Run/open extension
- Log into premium account
- Open Panel
- **Expected:** The panel should include the stats dashboard (See screenshot)

## Screenshots

Expected panel: 
image

### DIFF
--- a/src/js/popup/popup.js
+++ b/src/js/popup/popup.js
@@ -155,14 +155,13 @@ const serverStoragePanel = {
 
 async function choosePanel(numRemaining, panelId, premium, premiumSubdomainSet){
   const premiumPanelWrapper = document.querySelector(".premium-wrapper");
-  const premiumCountryAvailability = (await browser.storage.local.get("premiumCountries"))?.premiumCountries;
-
+  
   // Turned off label sync prompt for premium release
   // const shouldShowServerStoragePromptPanel = await serverStoragePanel.isRelevant();
   // if (shouldShowServerStoragePromptPanel) {
   //   serverStoragePanel.init(premium);
   // } else 
-  if (premium && premiumCountryAvailability?.premium_available_in_country === true) {
+  if (premium) {
     document.getElementsByClassName("content-wrapper")[0].remove();
     premiumPanelWrapper.classList.remove("is-hidden");
     premiumPanelWrapper


codemist (approved):
Checked this branch out, works as expected 👍

src/js/popup/popup.js
```diff
@@ -155,14 +155,13 @@ const serverStoragePanel = {
...
   const premiumPanelWrapper = document.querySelector(".premium-wrapper");
-  const premiumCountryAvailability = (await browser.storage.local.get("premiumCountries"))?.premiumCountries;
-
+  
   // Turned off label sync prompt for premium release
   // const shouldShowServerStoragePromptPanel = await serverStoragePanel.isRelevant();
   // if (shouldShowServerStoragePromptPanel) {
   //   serverStoragePanel.init(premium);
   // } else 
-  if (premium && premiumCountryAvailability?.premium_available_in_country === true) {
```
maxxcrawford, line 165: @Vinnl Was there a specific use-case for limiting the premium panel to only Premium country-available users?

Vinnl: tl;dr No, the country availability check can (and should) be removed.

Looking at the PR that introduced it (#246), I simply went by the name of the function I replaced (`premiumFeaturesAvailable`) and replaced that with this check. Of course, whether the Premium feature flag was enabled is actually something different than whether Premium is available in the user's current country, and in this case, if the user already has Premium even if they're in a different country where Premium is not available, it should still be shown.

Vinnl: I just went over other instances of this check, but those refer to the upgrade prompts ("Get unlimited aliases"), so this is the only one that should be removed :)

maxxcrawford: @Vinnl Thanks for the background/double-check!

PR #259, opened 2022-01-20
Don't keep the first run tab open when logging in

Vinnl:
In the "First run" tab, if the user clicks "Sign Up / In", a new
tab used to be opened in which they could sign in. After signing
up/in, they would be redirected back to the dashboard, but in
another tab, the "first run" screen would sill be open and asking
them to sign in, which was weird. This instead starts the sign in
flow in the same tab.

(It's [been there since the beginning](https://github.com/mozilla/fx-private-relay-add-on/blame/5e1829b03b25eeda63a5b625dbd52aa875c162e4/extension/js/first-run.js#L32), so I don't think it was too conscious a decision to open in a new tab.)

### DIFF
--- a/src/js/first-run/first-run.js
+++ b/src/js/first-run/first-run.js
@@ -17,7 +17,7 @@ document.addEventListener("DOMContentLoaded", async () => {
     el.addEventListener("click", (e) => {
       e.preventDefault();
       sendRelayEvent("First Run", "click", e.target.dataset.eventLabel);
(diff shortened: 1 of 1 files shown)


maxxcrawford (approved):
[THANK YOU FOR FIXING THIS](https://media.giphy.com/media/S7Q8eVYii3z1Ak0N5B/giphy.gif)! This was a _Bug That Bugged ME™_.

Works in both Firefox and Chrome builds!

Vinnl:
Ha, I assume you picked the Dutch football team on purpose? That's some on-point gif-searching skills you've got there :)

PR #267, opened 2022-01-26
Also build preview of the Chromium extension in CI

Vinnl:
It's downloadable from the same place as the Firefox extension.


maxxcrawford (approved):
This looks great. Can you confirm that this artifact is configured for stage at this step?

Vinnl:
@maxxcrawford Yes it should be, that's done [a couple lines up](https://github.com/mozilla/fx-private-relay-add-on/blob/68d32b552194a0925cdd0cfaedd06b886e44096f/.github/workflows/sign-and-release-to-github.yml#L67-L68)! And in fact you can verify that right away by looking at [the built artifacts for this PR](https://github.com/mozilla/fx-private-relay-add-on/actions/runs/1752823975), where I see at least a reference to the stage site in the `manifest.json` of the Chrome build.

PR #268, opened 2022-01-26
New auth mpp 1531

groovecoder:
I want to get this up for review & comments. Lots of todos still ...

## TODOs
* Extension: Update all server requests to use new auth token.
  * Note: The server requests seem a bit messy, maybe this is the time to add a new `src/js/shared/api.js` module?
* Extension: Automatically refresh the access token when/before it expires.
* Server: Cache the token verification check so we don't hit FXA on every client request.


src/js/first-run/first-run.js
```diff
@@ -1,7 +1,35 @@
...
+   */
+  var text = "";
+  var possible = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+
+  for (var i = 0; i < Math.random() * (128 - 43) + 43; i++) {
+    text += possible.charAt(Math.floor(Math.random() * possible.length));
+  }
+
+  return text;
+}
```
Vinnl, line 17: Does "random" here mean "cryptographically secure random"? Because if so, `Math.random` doesn't provide that. (I'm a bit nervous about not using a library for authn algorithms anyway...)

And a nit to avoid var and use consistent capitalisation:

```suggestion
function generatePkceCodeVerifierValue() {
  /*
   * A code verifier is a random string using characters of
   * [A-Z] / [a-z] / [0-9] / "-" / "." / "_" / "~",
   * with a minimum length of 43 characters and a maximum length of 128 characters.
   */
  let text = "";
  const possible = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";

  for (let i = 0; i < Math.random() * (128 - 43) + 43; i++) {
    text += possible.charAt(Math.floor(Math.random() * possible.length));
  }

  return text;
}
```

Vinnl (commented):
Given that there's still a bunch of things to do, I don't see any major issues apart from feeling uneasy about implementing PKCE ourselves.

>     Note: The server requests seem a bit messy, maybe this is the time to add a new src/js/shared/api.js module?

@maxxcrawford You said something about something in our add-on submission not allowing for a bundler - could you clarify what that was? It would indeed be lovely to be able to just `import` a module and being able to re-use that.

src/js/first-run/first-run.js
groovecoder: How does https://www.npmjs.com/package/pkce-challenge look?

Vinnl: I don't have experience with any, unfortunately. pkce-challenge looks like it's made for Node.js, so probably won't work in the add-on. Checking with a former colleague, PKCE is limited enough that we might be able to get away with implementing it ourselves, so we could maybe just use [`window.crypto.getRandomValues`](https://developer.mozilla.org/docs/Web/API/Crypto/getRandomValues).

maxxcrawford: RE: Bundler question from a month ago: There's an extra submission/review step if we use a bundler. That may not be as heavy as a lift, but every add-on I've worked on at Mozilla, we didn't have one. 

https://extensionworkshop.com/documentation/publish/source-code-submission/

src/js/background/background.js
```diff
@@ -311,6 +313,12 @@ browser.runtime.onMessage.addListener(async (m) => {
     case "updateAddOnAuthStatus":
       await updateAddOnAuthStatus(m.status);
       break;
+    case "loadAliasesFromServerIntoStorage":
+      console.log('loading aliases from relay server ...');
```
maxxcrawford, line 317: nit: remove console log
```suggestion
```

src/js/background/background.js
```diff
@@ -311,6 +313,12 @@ browser.runtime.onMessage.addListener(async (m) => {
...
+      relayAddresses = await getAliasesFromServer('GET', null, {auth: true});
+      console.log(relayAddresses);
```
maxxcrawford, line 319: nit: remove console log
```suggestion
```

src/js/background/background.js
```diff
@@ -9,8 +9,8 @@ browser.runtime.onInstalled.addListener(async () => {
   if (firstRunShown) {
     return;
   }
-  const userApiToken = await browser.storage.local.get("apiToken");
```
maxxcrawford, line 12: Question: Do we need to do any sort of cleanup/removal/migration for the previous `userApiToken` storage item? Will users be logged out of the Relay add-on when this lands? Do we need to ship something to the website first?

src/js/first-run/first-run.js
maxxcrawford: TIL about `window.crypto.getRandomValues`!

src/js/first-run/first-run.js
```diff
@@ -1,7 +1,35 @@
...
+  );
+  const b64EncodedChallenge = btoa(String.fromCharCode(...new Uint8Array(pkceCodeChallengeDigest)))
+  return b64EncodedChallenge.replace(/\+/g, '-').replace(/\//g, '_').replace(/=/g, '');
+}
+
+
 document.addEventListener("DOMContentLoaded", async () => {
   const { relaySiteOrigin } = await browser.storage.local.get("relaySiteOrigin");
+  const fxaAuthBaseUrl = 'https://oauth.stage.mozaws.net/v1';
+  const clientID = '0d134b18b51f173e';
```
maxxcrawford, line 32: Suggestion (non-blocking): Set this as a variable at the top of the file. 

Note: I know we don't have an actual .ENV file setup, but I like to surface these where we can. 

```suggestion
  const FXA_CLIENT_ID = '0d134b18b51f173e';
```

src/js/first-run/first-run.js
```diff
@@ -12,10 +40,55 @@ document.addEventListener("DOMContentLoaded", async () => {
...
+      const authUrl = `${fxaAuthBaseUrl}/authorization`;
+      const scope = 'profile';
+      const accessType = 'offline';
+      const redirectUrl = browser.identity.getRedirectURL();
+      const state = Math.random().toString(36).substr(2);
+      const pkceCodeVerifier = generatePKCECodeVerifierValue();
+      console.log(`pkceCodeVerifier: ${pkceCodeVerifier}`);
+      const pkceCodeChallenge = await pkceCodeChallengeB64(pkceCodeVerifier);
+      console.log(`pkceCodeCallenge: ${pkceCodeChallenge}`);
+      const fullAuthUrl = `${authUrl}?client_id=${clientID}&scope=${scope}&access_type=${accessType}&redirect_uri=${redirectUrl}&state=${state}&code_challenge_method=S256&code_challenge=${pkceCodeChallenge}&response_type=code`;
```
maxxcrawford, line 55: Suggestion: Wrap this entire set of declarations into a helper function (not inside this click function)? (Eg: `buildFullAuthUrl()`, etc.)

src/js/popup/popup.js
```diff
@@ -429,7 +429,14 @@ async function popup() {
...
   document.querySelectorAll(".login-link").forEach(loginLink => {
-    loginLink.href = `${relaySiteOrigin}/accounts/profile?utm_source=fx-relay-addon&utm_medium=popup&utm_content=popup-continue-btn`;
+    loginLink.addEventListener("click", (e) => {
+      e.preventDefault();
+      sendRelayEvent("Login link", "click", e.target.dataset.eventLabel);
+      const authResult = browser.identity.launchWebAuthFlow({
+        url: 'https://oauth.accounts.firefox.com/v1/authorization'
+      });
+    });
+    // loginLink.href = `${relaySiteOrigin}/accounts/profile?utm_source=fx-relay-addon&utm_medium=popup&utm_content=popup-continue-btn`;
```
maxxcrawford, line 439: nit: remove commented code
```suggestion
```

src/manifest.json
```diff
@@ -69,6 +69,7 @@
 
   "permissions": [
       "<all_urls>",
+      "identity",
```
maxxcrawford, line 72: Question: Will this trigger a permission update? 

Also, it looks like it is, but I want to confirm cross-compat with Chrome: https://developer.chrome.com/docs/extensions/reference/identity/

maxxcrawford (changes requested):
I have a few larger concerns around retention/friction points for current users (and Chrome): 
- How we migrate current logged in users 
- How the new permission/flow works in Chrome
- If the new permission requires opt-in/approval from current users

PR #273, opened 2022-02-01
HOTFIX: Update CSS to target only Relay add-on elements base font size 

### LINKED ISSUE #272: Firefox Relay extension breaks the display of some websites (changes "zoom" level?)
Hi,

Installing/enabling the Firefox Replay extension seems to "zoom" the content of some websites.

Observed on macOS 11.6.3 with Firefox 96.0.3 and Firefox Developer Edition 97.0b9.

Steps for reproduction:

- Launch Firefox Developer Edition 97.0b9 with default blank profile (no extensions)
- Open https://www.lemonde.fr/
- The content is correctly displayed without any "zoom" level:

Capture d’écran 2022-02-01 à 19 47 31

- Install de Firefox Relay extension
- Return to the tab containing lemonde.fr
- The content is now "zoomed" by default:

Capture d’écran 2022-02-01 à 19 47 52

if I disable the extension, then the content reverts back to the initial and default size.

However, it does not seems to happen on all websites. Here is another website that exhibit this behavior: https://www.actu.fr/

maxxcrawford:
Fixes #272

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fx-private-relay-add-on",
-  "version": "2.2.1",
+  "version": "2.2.2",
   "description": "Firefox Relay",
   "main": "index.js",
   "dependencies": {},
--- a/src/css/relay.css
+++ b/src/css/relay.css
@@ -1,7 +1,3 @@
-:root, body {
-  font-size: 16px;
-}
-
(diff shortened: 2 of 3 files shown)


groovecoder (approved):
Code and spot-check of fix looks good.

PR #286, opened 2022-02-25
Show panels based on what your browser can supprt

### LINKED ISSUE #285: Hide "Sign back in with Relay" Annoucement on non-Firefox builds

maxxcrawford:
Fix #285 - Revise panel logic to show announcement news based on which features your browser supports

## Testing

_Testing note: Test both free and premium accounts._ 

Test on Chrome: 
- There should only be two announcement panels 

Test on Firefox: 
- There should only be three announcement panels

### DIFF
--- a/src/css/popup.css
+++ b/src/css/popup.css
@@ -515,6 +515,7 @@ onboarding-panel {
   line-height: 120%;
 }
 
+.is-last-panel .next-panel,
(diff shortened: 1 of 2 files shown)


codemist (approved):
Works as expected!! Thank you for picking this up.

PR #290, opened 2022-03-02
Remove attachment size limit panel

codemist:
**Firefox:**
- Free
https://user-images.githubusercontent.com/13066134/156411506-d2ff6706-8d8b-4e87-8b1e-c9d80944faf0.mov

- Premium
https://user-images.githubusercontent.com/13066134/156411941-2efe9c32-851a-47c8-af22-5a666ae5d619.mov

**Chrome:**
- Free
image

- Premium
image

I can be in charge of reversing these changes for the later release when there's more clarity with attachment size limits.

### DIFF
--- a/src/css/popup.css
+++ b/src/css/popup.css
@@ -516,7 +516,7 @@ onboarding-panel {
 }
 
 .is-last-panel .next-panel,
-#educationalAttachmentSizeLimit .previous-panel,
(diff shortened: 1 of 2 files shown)


maxxcrawford (approved):
I'm in favor of leaving the comments, as we know we're going to turn around and end up enabling at some point in the future. Nice, quick fix!

PR #301, opened 2022-03-22
Send browser vendor with telemetry events

maxxcrawford:
See https://github.com/mozilla/fx-private-relay/pull/1670

This PR adds sets a custom dimension (`dimension5`) for which browser vendor the add-on uses: Firefox or Chrome.


Vinnl (approved):
TIL about our metrics setup!

PR #319, opened 2022-04-28
Ensure the First Run shown is not shown on updates

Vinnl:
If `firstRunShown` somehow gets removed from the user's storage,
this still avoids showing the "first run" screen when the extension
was merely updated (or one of the other non-extension-install
events [1] trigger the `onInstalled` callback.

[1] https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/OnInstalledReason


maxxcrawford (approved):
Works for me! I wasn't able to run an upgrade to confirm, but it did the first run (on install) without issue. :)

PR #320, opened 2022-05-02
Remove code made redundant by the React UI

Vinnl:
This mainly pertains to mask labels: rather than scraping the
website UI for available masks and their labels, and
monkey-patching the DOM to add a label editor, we now communicate
explicitly with the website via the API (to get mask data) and the
custom element (to exchange local label data).


maxxcrawford:
Quick question before reviewing: Does this still offer support for users who don't sync labels with the server?

Vinnl:
@maxxcrawford Yep! The add-on [injects local labels into the website](https://github.com/mozilla/fx-private-relay-add-on/blob/MPP-1575-remove-redundant-ui/src/js/relay.firefox.com/installation_indicator.js#L15-L25), which will then show those to users, allows them to edit them, and when edited, [sends the new labels to the add-on](https://github.com/mozilla/fx-private-relay-add-on/blob/MPP-1575-remove-redundant-ui/src/js/relay.firefox.com/get_profile_data.js#L210-L228) (from [here](https://github.com/mozilla/fx-private-relay/blob/4bed7098d5d43359a387124403f57a3f92a4c3b7/frontend/src/hooks/localLabels.ts#L53-L57)). That means all the old infrastructure for local labels is no longer needed, which this PR removes.

maxxcrawford (approved):
🧹

PR #327, opened 2022-05-04
WIP: Refactor how the add-on detects if a user is logged into Relay or not

maxxcrawford:
Summary: There's a few seconds (race condition) when the dashboard is page is open that sends a signal to the add-on that the user is logged out. 

Note: This PR is not the root solve for the issue, but does patch the issue some.


maxxcrawford:
Closing in favor of https://github.com/mozilla/fx-private-relay-add-on/pull/332

PR #330, opened 2022-05-05
WIP/Spike: Web-compat Input Field Issue

maxxcrawford:
https://mozilla-hub.atlassian.net/browse/MPP-1803


maxxcrawford:
This is a working PR to talk through solutions. 

Testing on https://myaccount.nytimes.com/auth/login. 

The input for is ignoring all value modifications from outside of a user authentically typing on a keyboard. 

HOWEVER, 1Password is able to do it. I'm not sure if we have to maintain focus on the field while dispatching the events?

maxxcrawford:
Fixed by #359

PR #332, opened 2022-05-06
More reliably track user auth status

### LINKED ISSUE #314: Clicking on the “Manage all masks” link in the Field Icon menu makes the add-on think the user has logged out of their Relay account
#### Build:
- Stage;

#### Affected Platforms:
- Windows 10;

#### Browsers tested on:
- Firefox, Release version;
- Chrome, Release version;

#### Prerequisites:
- Be signed in into a Firefox Relay account;
- Have at least one alias created;
- Have the latest Stage version of the add-on installed;
- Have the Relay Field Icon displayed;

#### Steps to reproduce:
1. Go to any site that displays a fill-able email field;
2. Click on the Relay Field Icon;
3. Click on the “Manage all masks” link;
4. Open the Add-on toolbar menu and observe it;
5. Go back to the fill-able email field;
6. Click on the Relay Field Icon and observe its menu;

#### Expected result:
- The menus of the Add-on should be displayed correctly to the logged-in user;

#### Actual result:
- After clicking the “Manage all masks” link, the Add-on menus look like the user is logged off;

#### Notes:
- There is a brief moment after clicking the “Manage all masks” link where the Relay Homepage is visible, and on it there is no user logged in;
- This problem can be fixed by refreshing the Relay dashboard or by clicking on the “Continue” button inside the “not-signed-in” view of the Add-on toolbar menu;
- See attached media:
logout

Vinnl:
Previously, we'd detect whether the user was logged out by looking
at the `<firefox-private-relay-addon>` custom element and
inspecting its `data-user-logged-in` property. However, since the
React website checks the user's logged-in status _after_ rendering,
this could result in us concluding that the user was logged out
before it had been able to check, and logging the user out in the
add-on.

Additionally, it overloaded the API between the add-on and the
website, which has also caused a bug in the past: in addition to
the add-on injecting data into the website using attributes of
`<firefox-private-relay-addon>`, _some_ attributes were instead
used for communication the other way around, with the website
sending data to the add-on (in this case, the user's auth status).
Whereas generally, we use the `<firefox-private-relay-addon-data>`
element for communication from the website to the add-on.

With this change, we check whether the user is logged in in a
different way: if the user visits the Relay homepage (e.g. after
hitting the logout button) and does *not* get redirected to
/accounts/profile/, we can assume that they're logged out, and thus
will clear their locally-cached data.

Otherwise, if they are redirected to `/accounts/profile`, and we
were able to successfully read their data, we can conclude that
they're now logged in (so we also call `updateAddOnAuthStatus`,
although we don't actually do anything in response yet).

(Note that this does depend on https://github.com/mozilla/fx-private-relay/pull/1816 working correctly; I'm currently not seeing that work on production, even though it works on stage, and I'm [not sure why yet](https://mozilla.slack.com/archives/C013CSYEL5T/p1651846196483509).)

This solves both the bug, and makes website-add-on-communication
more consistent. When this is released, we can remove [these lines](https://github.com/mozilla/fx-private-relay/blob/8060fc4eb823de715323780c1918e8ff0a49f87c/frontend/src/pages/_app.page.tsx#L93-L97).

Fixes #314, supersedes #327.

### DIFF
--- a/src/js/background/background.js
+++ b/src/js/background/background.js
@@ -294,7 +294,7 @@ async function makeRelayAddress(description = null) {
 async function updateAddOnAuthStatus(status) {
   // If user is no longer logged in, remove the apiToken attribute. 
   // This will cause the "Sign in" panel to be visible when the popup is opened.
-  if (status === "False") {
+  if (status === false) {
     await browser.storage.local.remove("apiToken");
   }
 }
--- a/src/js/relay.firefox.com/get_profile_data.js
+++ b/src/js/relay.firefox.com/get_profile_data.js
@@ -319,6 +319,10 @@
     });
 
     await browser.runtime.sendMessage({
+      method: "updateAddOnAuthStatus",
+      status: true,
+    });
+    await browser.runtime.sendMessage({
       method: "rebuildContextMenuUpgrade",
     });
   }
--- a/src/js/relay.firefox.com/inject_addon_data.js
+++ b/src/js/relay.firefox.com/inject_addon_data.js
@@ -29,22 +29,4 @@
       el.dataset.localLabels = JSON.stringify(localLabels);
     }
   });
-
-  // Check for <firefox-private-relay-addon>
-  const addonDataElement = document.querySelector(
-    "firefox-private-relay-addon"
-  );
-  if (addonDataElement) {
-    // Query if the user is logged in or out
-    const isLoggedIn = document.querySelector("firefox-private-relay-addon")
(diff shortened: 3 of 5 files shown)


maxxcrawford (approved):
So clean! Confirmed log out/in worked without issue. Thanks for the root fix here. 😎

Vinnl:
Blocker has been resolved by changing the cache headers, so this can be merged!

Issue #331, opened 2022-05-05
Relay add-on doesn't save the site origin anymore


Bianca-O:
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit

**Browsers tested on:**
Firefox, latest release
Chrome, latest release

**Prerequisites:**
The Relay extension is installed in the browser.
The Data collection is turned off.

**Steps to reproduce:**

1. Navigate to Relay page and log in.
2. Go to a different website that has an email field and click the Relay icon in the email field.
3. Generate new mask.
4. Go to Relay dashboard and refresh the page.
5. Observe the generated alias and it's label.

**Actual result:**
The label is initially saved for the generated alias and after refreshing the page it is not displayed anymore.

**Expected result:**
The label is still displayed even after refreshing the page.

**Notes:**
Another thing I noticed is that in the field icon menu the label for site generated aliases is also not displayed. I'm not sure if this is an issue or not.

![addon doesnt save site origin](https://user-images.githubusercontent.com/79697848/166954672-56efec1a-2535-49b2-bc6d-fa997dcafc8d.gif)

![addon issue](https://user-images.githubusercontent.com/79697848/166954835-d6ecc925-3159-4500-ac7d-720e73eb7ef6.gif)

PR #336, opened 2022-05-18
Alias list listener

Vinnl:
This is a companion PR to a website one *that should be deployed before this is released*: https://github.com/mozilla/fx-private-relay/pull/1962

The website added some more events that the add-on can respond to.
Specifically, we now ensure that our local cache of data is updated
whenever the user registers their subdomain on the website, or adds
or removes a mask on the website. We also now refresh non-settings
pages whenever server-side storage of labels is toggled based on a
website event.

The three features can be tested separately, with the website changes checked out as well.

1. Open the pop-up, note the mask count. Create a mask on the website, open the pop-up again. The count should have updated (it didn't before).
2. Like above, but deleting a mask.
3. Register a subdomain. Check the extension storage, note that `premiumSubdomainSet` is properly updated (it wasn't before).
4. Open two website pages, and toggle server-side label storage on one of them. Notice that the other refreshes when this is done.


src/js/relay.firefox.com/get_profile_data.js
```diff
@@ -87,38 +104,59 @@
...
+        });
+
+        return {
+          ...address,
+          description: matchingLocalAddress?.description ?? address.description,
+          generated_for: matchingLocalAddress?.generated_for ?? address.generated_for,
+          used_on: matchingLocalAddress?.used_on ?? address.used_on,
+        };
+      });
+    }
```
Vinnl, line 152: Note: we'll probably want to call this function on [`masks` here](https://github.com/mozilla/fx-private-relay-add-on/blob/7f1653ec82a60cd35d68f4bfacf0fe7692a3c7a5/src/js/background/background.js#L63) as well, to solve https://github.com/mozilla/fx-private-relay-add-on/issues/331.

src/js/background/background.js
```diff
@@ -268,8 +262,6 @@ async function makeRelayAddress(description = null) {
   ]);
   browser.storage.local.set({ relayAddresses: updatedLocalRelayAddresses });
 
-  await refreshAccountPages();
-
```
maxxcrawford, line 272: We still need a refresh mechanism to update the dashboard (if open) when the add-on generates a new mask. I want to flag this for a future enhancement.

Vinnl: Yeah that sounds good. I don't think we used to have that before either, did we? Otherwise we can just re-use that mechanism (i.e. stick with refreshing until we have proper add-on-website comms).

PR #334, opened 2022-05-12
Feature: Add "Used on Current Site" List of Masks to in-page menu

maxxcrawford:
## Summary

[MPP-1638](https://mozilla-hub.atlassian.net/browse/MPP-1638) 

Blocked (Cannot submit to AMO until) by: 
- https://github.com/mozilla/fx-private-relay/pull/1968
- https://github.com/mozilla/fx-private-relay/pull/1969

This PR adds additional functionality to the in-page menu. Now, when users fill in email inputs using the add-on, we will remember that and service it the next time they're on the site. 

## Caveats

- ~Local storage only (not syncing with the server) has not been tested/is not currently supported~
- ~Website limitation:  `domainaddresses` does not yet support the `used_on` field~
- This functionality does not apply to context-menu/alt-clicking
  - Draft started https://github.com/mozilla/fx-private-relay-add-on/pull/342

### Free account: 

- Start with an empty account (with sync enabled)
- Create a few masks from the dashboard page
  - _Optional: Set a label_
- Go to a website with an email input form
- Click on the Relay icon
- **Expected:** The in-page menu should open up with the masks in the "Select from your current email masks" list
  - _Note: If you set a label, it should show the label in this list, rather than the mask address._
- Click on a mask
- **Expected:** The mask address should be auto-filled into the email input
- Click on the Relay icon again
- **Expected:** The in-page menu should open up with that selected mask now in the "Previously used on this website" list 
  - _Note: If you created multiple masks, there will now be two lists: "Select from your current email masks" and "Previously used on this website"_
- Click on "Generate new mask" button
- **Expected:** A mask address should be created and then auto-filled into the email input
- Click on the Relay icon again
- **Expected:** There should be two items in the "Previously used on this website" list 
  - _Note: The newest mask address should have a label of the current website you're on._

### Premium Account

- Start with a premium, empty account (with sync enabled)
- Create a few masks from the dashboard page
  - _Note: If you create a custom mask (with a registered subdomain), this will NOT work until https://github.com/mozilla/fx-private-relay/pull/1968 is merged. You can still re-use these with the add-on, but it will not save/store site usage._
- Go to a website with an email input form
- **Expected:** The in-page menu should open up with the masks from a single (unlabeled) list. 
  - _Optional: If you created over five masks, you should have a search field to search that visible list of masks._
- Click on a mask
- **Expected:** The mask address should be auto-filled into the email input
- Click on the Relay icon again
[...]

PR #342, opened 2022-05-23
Feature: Store/show masks used on specific sites in the context menu

maxxcrawford:
Follow-up feature for https://github.com/mozilla/fx-private-relay-add-on/pull/334

## Testing

I've outlined how to test this feature stand-alone style. However, I recommend to also mix it up with a combination of alt/right clicking AND clicking the Firefox Relay icon (in-page menu). See the previous PR for steps to test that. 

### Free account: 

- Start with an empty account (with sync enabled)
- Create two-three masks from the dashboard page
  - _Optional: Set a label_
- Go to a website with an email input form (Example: Facebook.com) 
- Alt-click (right-click) on an input field
- **Expected:** The content menu should open, with a Firefox Relay menu item (See example screenshot)
  - There should be the following menu items: Generate new mask, Recent masks, Manage all masks, Get unlimited masks
  - _Note: If you set a label, it should show the label in the "Recent masks" list, rather than the mask address._
- Click on a mask in the "Recent masks" list
- **Expected:** The mask address should be auto-filled into the email input
- Alt-click (right-click) on an input field again
- **Expected:** The content menu should open, with a Firefox Relay menu item, with a new menu entry: "Use existing mask from this site..."
- Click on "Generate new mask" button
- **Expected:** A mask address should be created and then auto-filled into the email input
- Alt-click (right-click) on an input field again
- **Expected:** There should be two items in the "Use existing mask from this site..." list 
  - _Note: The newest mask address should have a label of the current website you're on._

### Premium Account

The behavior should be the same for premium accounts, with two additional notes: 
- You can also use custom domain created masks "github@example.mozmail.com"
- There should be no "Get unlimited masks" upgrade menu item 

## Screenshots

image

image

image

### DIFF
--- a/src/js/background/context-menu.js
+++ b/src/js/background/context-menu.js
@@ -1,9 +1,9 @@
 /* global patchMaskInfo */
 
-// The static data used to create different context menu items. 
+// The static data used to create different context menu items.
 // These are the same everytime, as opposed to the dynamic menu items: reusing aliases
 // See these docs to better understead the context menu paramaters
-// https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus/create#parameters 
+// https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus/create#parameters
 const staticMenuData = {
   existingAlias: {
     type: "radio",
@@ -44,17 +44,19 @@ const staticMenuData = {
   },
   useExistingAliasFromWebsite: {
     id: "fx-private-relay-use-existing-aliases-from-this-site",
-    title: browser.i18n.getMessage("pageInputIconUseExistingAliasFromTheSite_mask"),
+    title: browser.i18n.getMessage(
+      "pageInputIconUseExistingAliasFromTheSite_mask"
+    ),
     visible: true,
     contexts: ["all"],
-  }, 
+  },
   useExistingAlias: {
     id: "fx-private-relay-use-existing-aliases",
     title: browser.i18n.getMessage("pageInputIconRecentAliases_mask"),
     visible: true,
     contexts: ["all"],
-  }
-}
+  },
+};
 
 // Existing Relay/random aliases will get their own context menu items,
 // identified by the following prefix followed by their alias ID:
@@ -79,19 +81,27 @@ async function getCachedServerStoragePref() {
 // This object is defined as global in the ESLint config _because_ it is created here:
 // eslint-disable-next-line no-redeclare
 const relayContextMenus = {
-  init: async (currentWebsite=null) => {
-    
+  init: async (currentWebsite = null) => {
     if (!browser.contextMenus) {
       throw new Error(`Cannot create browser menus`);
     }
 
     // Remove the listener so we don't add the same one multiple times
-    if (browser.storage.onChanged.hasListener(relayContextMenus.listeners.onLocalStorageChange)) {
-      await browser.storage.onChanged.removeListener(relayContextMenus.listeners.onLocalStorageChange);
+    if (
+      browser.storage.onChanged.hasListener(
+        relayContextMenus.listeners.onLocalStorageChange
+      )
+    ) {
+      await browser.storage.onChanged.removeListener(
+        relayContextMenus.listeners.onLocalStorageChange
+      );
     }
 
     const userApiToken = await browser.storage.local.get("apiToken");
-    const apiKeyInStorage = Object.prototype.hasOwnProperty.call(userApiToken, "apiToken");
+    const apiKeyInStorage = Object.prototype.hasOwnProperty.call(
+      userApiToken,
+      "apiToken"
+    );
 
     if (!apiKeyInStorage) {
       // User is not logged in. Do not do anything.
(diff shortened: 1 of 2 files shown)


maxxcrawford (commented):
Self-review

src/js/background/context-menu.js
```diff
@@ -86,65 +113,87 @@ const relayContextMenus = {
...
-          createExistingAliases: true,
-          parentMenu: staticMenuData.useExistingAlias,
-          exisitingSite: false,
-          currentWebsite
-        }, aliases)
+      // Create Use Existing Alias submenu
+      if (
+        currentWebsite &&
+        masksWereGeneratedOrUsedOnCurrentWebsite &&
+        userHasSomeAliasesCreated
```
lloan, line 139: praise: thank you for creating human-readable variables! Makes it so much easier to understand what is going on.

src/js/other-websites/inpage_menu.js
```diff
@@ -311,32 +323,35 @@ function hasMaskBeenUsedOnCurrentSite(mask, domain) {
...
 
 function haveMasksBeenUsedOnCurrentSite(masks, domain) {
-  return masks.some(mask => {
-    
+  return masks.some((mask) => {
     const domainList = mask.used_on;
 
     // Short circuit out if there's no used_on entry
-    if (domainList === null || domainList === "" ||  domainList === undefined) { return false; }
+    if (domainList === null || domainList === "" || domainList === undefined) {
```
lloan, line 344: suggestion (non-blocking): I've seen this logic at least a handful of times - I wonder if we should have a function that does this check and reuse it.

lloan (approved):
I tested this and it works.

src/js/background/context-menu.js
```diff
@@ -153,15 +172,42 @@ const relayContextMenus = {
...
+    
+      const currentMaskType = selectedAliasObject[0].mask_type;
+      const currentUsedOnValue = selectedAliasObject[0].used_on;
+    
+      const currentPage = new URL(tab.url);
+      const currentPageHostName = currentPage.hostname;
+ 
+      // If the used_on field is blank, then just set it to the current page/hostname. Otherwise, add/check if domain exists in the field
+      const used_on = (currentUsedOnValue === null || currentUsedOnValue === "" ||  currentUsedOnValue === undefined)
+        ? `${currentPageHostName},` : relayContextMenus.utils.addUsedOnDomain(currentUsedOnValue, currentPageHostName);
```
Vinnl: Maybe the logic described in the comment could be moved into `addUsedOnDomain`? I think it'll be relevant every time it gets called, no?

src/js/background/context-menu.js
```diff
@@ -153,15 +172,42 @@ const relayContextMenus = {
...
+        auth:true,
+        mask_type: currentMaskType,
+      };
+
+      // Save what site this mask was used on before filling input
+      if (serverStoragePref) {
+        await patchMaskInfo("PATCH", parseInt(selectedAliasId, 10), data, options);
+      } else {
+        selectedAliasObject[0].used_on = used_on;
+        browser.storage.local.set({ relayAddresses: relayAddresses });
```
Vinnl, line 249: I might be missing a bit too much context in the PR review view, but `relayAddresses` doesn't seem to get changed? Are you sure there's a need here to save it to local storage again?

src/js/background/context-menu.js
```diff
@@ -300,10 +355,15 @@ const relayContextMenus = {
...
-      return relayAddresses.some(alias => website === alias.generated_for);
+    checkIfAnyMasksWereGeneratedOrUsedOnCurrentWebsite: async (website) => {
+      const relayAddresses = await relayContextMenus.utils.getAliases();
+
+      // Short circuit if not current site detected
+      if (website === null) {
+        return false;
+      }
+
+      return relayAddresses.some(alias => website === alias.generated_for || alias.used_on?.includes(website));
```
Vinnl: Probably an edge case, but wouldn't this potentially match e.g. a mask used on `github.io`, even though the current user is at `othersite.github.io`? We miight want to do exact matches:

```suggestion
      return relayAddresses.some(alias => website === alias.generated_for || alias.used_on?.split(",").includes(website));
```

(Likewise for the other instances of `includes` below.)

Vinnl (approved):
Not sure if I was supposed to review this as well, but since it was assigned to me... :) Just added a question about potential erroneous partial matches.

Thanks for splitting out the formatting changes into a separate commit!

src/js/background/context-menu.js
maxxcrawford: The `selectedAliasObject[0].used_on` is modified, which is part of `relayAddresses` object. We're making sure to save the new `used_on` entry for the selected/used mask. 

I confirmed this by outputting `relayAddresses` before and after the `used_on` setting on L207.

maxxcrawford: I'll add a comment to make this more visible!

src/js/background/context-menu.js
maxxcrawford: Ahhh. Great callout. Migrated logic (and renamed the function to `checkAndStoreUsedOnDomain()`) in both the context menu and the in-page menu scripts.

Revised in cf7b8922020cc1f389f341b503d3362b4cfa4132

src/js/background/context-menu.js
maxxcrawford: Per conversation on Slack with Tony: 

> Hey reviewing right now. and i think that's okay, can we try releasing it like this and if we get feedback that it's messing users up or causing confusion we can file a bug to change it?

Adding a comment to capture this bug for future us. Comment added in 146b305533e54ea8a749cab984c3f61f891ab852

maxxcrawford:
```
  // BUG: Using includes without spliting on ",", there can 
  // be some false positive results for entries that share the same root domain. 
  // eg: Github mask would show up on example.github.io
  // See full conversation: https://github.com/mozilla/fx-private-relay-add-on/pull/342#discussion_r897755698
```

PR #345, opened 2022-05-24
Fix continuous prerelease to GitHub

Vinnl:
Somewhere between 2022-05-12 [1] and 2022-05-16 [2], `web-ext sign`
(or the API behind it) started producing different filenames, which
meant that the artefact uploader could no longer find the output.

See the output of the sign jobs:

image

vs.

image

[1] https://github.com/mozilla/fx-private-relay-add-on/actions/runs/2310548121
[2] https://github.com/mozilla/fx-private-relay-add-on/actions/runs/2329060749

PR #346, opened 2022-05-24
Also deal with renamed artefacts

Vinnl:
Followup of https://github.com/mozilla/fx-private-relay-add-on/pull/345 - apparently the `.xpi` name is also different. @maxxcrawford any idea what could have caused this? Any risk of this reverting back to the old name later?


maxxcrawford (approved):
This one too. :)

Issue #108, opened 2021-08-04
Email alias disappears on Instagram when relay add-on is installed and relay icon activated


Bianca-O:
**Build:**
Stage

**Affected platforms:**
Windows 10 64-bit

**Browsers tested on:**
Firefox

**Preconditions:**
The user is logged in to Firefox Relay page.
Relay add-on is installed.
The user is logged out of Instagram.

**Steps to reproduce:**

1. Navigate to https://www.instagram.com
2. Click on Sign Up button.
3. Refresh the page.
4. Click the Relay icon in the email field and generate a new alias.
5. Click in the other fields (e.g. Full Name).
2. Observe the email field.

**Actual result:**
The generated email alias disappears from the email field.
 
**Expected result:** 
The generated email alias is still displayed in the email field.
![instagram - relay add-on issue1](https://user-images.githubusercontent.com/79697848/128165657-fa4751c1-3390-4dad-822e-35f8bf556884.gif)

PR #359, opened 2022-06-03
Fixing web combat disappearing input issue

codemist:
When generating an email mask the following websites, the add-on then click away from that field the email mask clears from the text field. This PR should solve that bug by setting the `bubbles` property to true in the `Event` construction.

Preview:

https://user-images.githubusercontent.com/13066134/171930577-27db4354-db3e-45eb-a29e-8effe51a92f4.mov

List of sites the bug occurs on for testing: 
[New York Times](https://www.nytimes.com/)
[Zazzle](http://zazzle.com/) account registration 
[Ebay](https://www.ebay.com/)
[Homedepot](http://homedepot.com/) account registration

### DIFF
--- a/src/js/other-websites/fill_relay_address.js
+++ b/src/js/other-websites/fill_relay_address.js
@@ -15,6 +15,7 @@ function fillInputWithAlias(emailInput, relayAlias) {
   // Set the value of the target field to the selected/generated mask
   emailInput.value = emailMask;
 
+
   emailInput.dispatchEvent(new Event('input', {bubbles:true}));
 
 }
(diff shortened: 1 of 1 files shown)


src/js/other-websites/fill_relay_address.js
```diff
@@ -14,13 +15,12 @@ function fillInputWithAlias(emailInput, relayAlias) {
   // Set the value of the target field to the selected/generated mask
   emailInput.value = emailMask;
 
-  emailInput.dispatchEvent(
-    new InputEvent("relay-address", {
-      data: emailMask,
-    })
-  );
+
+  emailInput.dispatchEvent(new Event('input', {bubbles:true}));
```
codemist, line 19: By default, events manually created and dispatched like this don't propagate (bubble) up the DOM tree like events normally do, so adding it back in more securely sets the `value` property of the `input` object.

groovecoder:
Spot-check LGTM for context menu. I wasn't able to get the in-field icon to show up on HomeDepot or Zazzle though?

codemist:
@groovecoder  Hmm.. that may just still be a pre-existing issue with some of the icons not showing up/being wonky on specific sites.

Looping @maxxcrawford in about that

src/js/other-websites/fill_relay_address.js
Vinnl: Wow, this is a great find - I'd not expected at all that manually-created events would not bubble! I expect this will solve issues at _many websites_ (#108, for one).

Question: do you know why there was a `{data: emailMask }` property before? It still appears to work, so I presume it wasn't necessary,

Vinnl (approved):
Really great find @codemist :clap: I can also confirm that the icon already doesn't appear for Zazzle and Home Depot in `main`, so that's not a regression introduced by this PR.

codemist:
@maxxcrawford Runs the same on Chrome!

PR #361, opened 2022-06-07
Remove transitionary "type" property on masks

Vinnl:
This property was preserved to ensure that if the add-on update
reached people before the website update, people's experience
wouldn't break. However, now that the website has been updated to
look for the mask_type property, we no longer need to inject the
redundant and deprecated type property.

This is the (independent) sister PR to https://github.com/mozilla/fx-private-relay/pull/2050 and a follow-up to https://github.com/mozilla/fx-private-relay-add-on/pull/344.


maxxcrawford (approved):
Good to go!

Issue #22, opened 2020-05-01
High z-index value for email field wrapper causes stacking issues


alexgibson:
The `relay-email-input-wrapper` element that gets added around email input fields has the following CSS z-index value:

```css
z-index: 99999999999999999;
```

I'm sure this property is likely needed for a reason, but its value is way too high. This results in the email field appearing visually stacked on top of pretty much any other element that may be on a website.

For example, sticky navigation:

![image](https://user-images.githubusercontent.com/400117/80793114-1dcfdc80-8b8e-11ea-9538-968cda6828d5.png)

lolrepeatlol:
I'm still experiencing the same issue on Firefox 95.0 and the latest version of Nightly (97.0, 2021-12-20) with this extension. 

https://user-images.githubusercontent.com/17863082/146827395-4a04feda-b6bb-44c8-9342-20e133120e38.mp4

Paratron:
This is still present today. This addon breaks pages - come on, this is a major issue, why is it lying around for so long!

jwhitlock:
Also happens on https://accounts.nintendo.com. The email input is displayed on-top of the CAPTCHA:

![accounts nintendo com](https://user-images.githubusercontent.com/286017/171215095-c5380ba9-1c40-4e60-8cc3-8f06b42578a1.png)

Does not occur if Relay extension is disabled (but 1password extension enabled)

reemhamz:
I created a PR to fix this lingering issue: https://github.com/mozilla/fx-private-relay-add-on/pull/367
Would love to have someone from the Relay team review sometime soon!

Paratron:
Two. Years. This is so embarrassing.

czlucius:
Hi @alexgibson 
I would like to comment on this, which I assume affects many people using Relay, including me.

I've looked at how other extensions do this, and one extension in particular, Kee (Password Manager) (source [here](https://github.com/kee-org/browser-addon)) does it in an interesting way so overlay issues don't really happen:
- Setting the background image of the input field to Relay's icon via style setters in JS
- Adjusting the size, position, etc to align at the right at an appropriate dimension.
- When the user hovers, it checks if it exceeds the threshold then based on that it shows the pointer icon, else the regular icon.
- Using the same check, display the overlay if user is detected clicking on the icon (based on threshold), else nothing is performed.

This should eliminate the need for the `z-index` as well as the input overlay.
[...]

PR #367, opened 2022-06-23
Reduced `z-index` on email field wrapper 

reemhamz:
## Problem
Users who had Relay enabled experienced issues of having the email input field stacking on top of almost any other element on the page. Here's an example: 
Screen Shot 2022-06-13 at 3 42 36 PM

## Original issue
https://github.com/mozilla/fx-private-relay-add-on/issues/22

## My solution
With this fix, I mainly just reduced the `z-index` from `9999999999999....` to a `10`. I tested this out and this is the result:
Screen Shot 2022-06-23 at 11 54 10 AM

Maybe this needs more testing, but the issue has been up for some years with no action, so thought I'd take a swing at it. I'm sure there was a reason the `z-index` was so high in the first place.

### DIFF
--- a/src/css/relay.css
+++ b/src/css/relay.css
@@ -41,7 +41,7 @@
   border: 0;
 }
 
-.fx-relay-menu-iframe iframe { 
+.fx-relay-menu-iframe iframe {
   border-radius: 4px;
   background: var(--relayGrey10);
   background-color: var(--relayGrey10);
@@ -50,7 +50,7 @@
   animation: fxRelayFadeIn 0.2s ease forwards;
 }
 
-.fx-relay-menu-iframe { 
+.fx-relay-menu-iframe {
   width: 320px;
   /* The top/right positioning here is necessary to align the box correctly (offset) underneath the "triangle" directly below the in-page Relay logo */
   top: 47px;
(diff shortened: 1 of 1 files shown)


lolrepeatlol:
I feel like this shouldn't be merged and we need an actual solution for this issue. It's still going to present the same problem for websites that use a z-index under 10 for other things.

reemhamz:
Yeah, it's not necessarily the most inventive solution. However this issue has been around for some years and I mainly wanted to bring eyes to the matter. I think this is a good start compared to its initial `z-index` value of 99999999...99..9

maxxcrawford:
This definitely is a known issue. 

It would make sense to do some sort of CSS query/lookup to see if there's already a set z-index value. 

I'll share this with the team! Thanks for taking a whack at this.

Shnoulle:
No : this not fix the issue : there are no reason to have a z-index.

This not fix the issue in mybb foruml when search for an user

Capture d’écran du 2022-09-22 15-09-41
With z-index to 10.

Removing the z-index is a lot more a better fix : https://github.com/mozilla/fx-private-relay/pull/632

janbrasna:
I believe this was superseded by #462

Vinnl:
Thank you @janbrasna, that is correct!

Issue #263, opened 2022-01-25
The “Firefox Relay” submenu from the context menu closes by itself when the user hovers over it after the first time they open the context menu


AOiegas:
#### Build:
- Stage;

#### Affected Platforms:
- Windows 10;

#### Browsers tested on:
- Firefox, latest version;

#### Prerequisites:
- Be signed in into a Firefox Relay account with or without an active subscription;
- Have the latest version of the Stage Firefox Relay extension installed;

#### Steps to reproduce:
1. Go to any page that has a fill-able email field;
2. Right click on the field, opening the context menu;
3. Hover over the “Firefox Relay” option;
4. Hover over a different option;
5. Hover again over the “Firefox Relay” option;
6. Observe the behavior;

#### Expected result:
- There are no visible problems in the interactions;

#### Actual result:
- When the user opens the context menu for the first time and hovers over the “Firefox Relay” option, the submenu closes by itself;

#### Notes:
- Hovering again on the option opens it without issues;
- This always happens the first time a user opens the submenu; 
- There do not seem to be different behaviors depending on the number or type of generated aliases; 
- See attached media:
![contextmenu2](https://user-images.githubusercontent.com/85748693/150964129-38194995-3742-4c05-94c2-fafd54df4712.gif)

lloan:
This is fixed by #369 

@maxxcrawford should we mark this one complete/done?

maxxcrawford:
@lloan Good call! 

Closed by #369

PR #372, opened 2022-06-24
MPP-1608 Context menu closes by itself 

lloan:
This PR will fix MPP-1608 - The “Firefox Relay” submenu from the context menu closes by itself when the user hovers over it after the first time they open the context menu.

Proposed solution: 
`getAliases` now returns masks from local storage instead of fetching from API.


lloan:
@maxxcrawford just realized I reviewed that PR that fixed this yesterday. #369 fixes this issue #263

PR #373, opened 2022-06-27
Only use exact matches when searching used_on

Vinnl:
Before, if the string with comma-separated domains in the used_on
field included the current domain, we'd conclude that the
respective mask was used on the current domain — even if there was
just a partial match on the domain. For example, if a mask had
`fox.com` in its used_on field, that mask would then be considered
used on `firefox.com` as well.

With this change, only an exact match on `firefox.com` will count
as being used on that site.

How to test: I didn't find a site to properly test it on (e.g. the hypothetical above doesn't work, both because `firefox.com` immediately redirects to mozilla.org, as well as because Fox's domain is actually `www.fox.com`. Thus, I just went to http://127.0.0.1:8000/admin/, updated an address's `used_on` field (I set it to `notwww.mozilla.org`), then went and visited a domain that partially matched that value (I visited www.mozilla.org - there's an email field at the bottom of the page).

Without this change, you should see the above mask being listed as used on that site in both the in-page menu, as well as the context menu. With this change, it should not be listed.

This fixes MPP-2073, and is a follow-up to https://github.com/mozilla/fx-private-relay-add-on/pull/373.

### DIFF
--- a/src/js/background/context-menu.js
+++ b/src/js/background/context-menu.js
@@ -382,13 +382,9 @@ const relayContextMenus = {
         return false;
       }
 
-      // BUG: Using includes without spliting on ",", there can 
-      // be some false positive results for entries that share the same root domain. 
-      // eg: Github mask would show up on example.github.io
-      // See full conversation: https://github.com/mozilla/fx-private-relay-add-on/pull/342#discussion_r897755698
       return relayAddresses.some(
         (alias) =>
-          website === alias.generated_for || alias.used_on?.includes(website)
+          website === alias.generated_for || alias.used_on?.split(",").includes(website)
       );
     },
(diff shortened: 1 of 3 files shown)


maxxcrawford (approved):
Easy fix! 

I ran into one issue: I was not able to get the context menu to show up when I first tested this branch locally. I restarted all the apps a few times (website, add-on) and finally got it to work. 

Can you confirm its working as normal for you?

Vinnl:
@maxxcrawford I did see the regular context menu? Can you reproduce the error in `main`?

maxxcrawford:
> 

No. I wasn't - which is why I was nervous. Let me confirm one more time locally.

maxxcrawford:
@Vinnl Confirmed! This is good to land IMO.

PR #374, opened 2022-06-30
Fix label data loss

Vinnl:
Fixes https://github.com/mozilla/fx-private-relay/issues/2098 / MPP-2097.

The fix is small, but took a lot of tracking down. To reproduce,
sign in with an account with server-side label storage enabled and
the add-on installed. Server-side storage can be enabled in `/accounts/settings/`:

image

Make sure there are masks with a label. Then
right-click on an email field; the function `getServerStoragePref`
in background.js should now be setting `server_storage: true` in
the extension's local storage.

Now disable server-side storage, and reload the Relay website.
Now, instead of storing the updated preferences in the correct
place in the extension's storing, it will store it as
`settings: { server_storage: false }`.

Now, in an email field, open the menu from the icon button. It will
now check the `server_storage` property in the extension storage
(which is incorrectly still set to `true`), then re-fetch the masks
from the server, thereby overwriting locally-stored labels.

With this fix, the `server_storage` pref will be stored correctly,
and thus locally-stored labels will no longer be overwritten.

### DIFF
--- a/src/js/relay.firefox.com/get_profile_data.js
+++ b/src/js/relay.firefox.com/get_profile_data.js
@@ -100,9 +100,7 @@
 
     browser.storage.local.set({
       profileID: parseInt(serverProfileData[0].id, 10),
-      settings: {
-        server_storage: serverProfileData[0].server_storage,
-      },
+      server_storage: serverProfileData[0].server_storage,
     });
 
     const siteStorageEnabled = serverProfileData[0].server_storage;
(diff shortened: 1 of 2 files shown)


maxxcrawford (commented):
Oh, boy. This is one of those "tiny, tiny fix" PRs that make are so hard to catch. Fantastic work! 

Leaving this to @codemist to review.

codemist (approved):
I came across some random bugs at times during testing where the labels would suddenly disappear when I would toggle server storage on and off, but they were always 'retrievable' after reloading a couple times and/or generating a new alias.

Vinnl:
Hmm, that doesn't sound good, though also potentially not related to this change? If I'm understanding it correctly, there was no data loss, is that correct? Assuming that's the case, I'll go ahead and merge this so it can get tested, and then when we're sure that's the case it can be released ASAP.

PR #382, opened 2022-07-08
Create production packages when pushing a tag

Vinnl:
This way, we should never have to manually (forget to) configure the add-on code for the production site or to build the extension artefacts for submission; we can just download the fully prepared package from GitHub Actions.

### DIFF
--- a/.github/workflows/sign-and-release-to-github.yml
+++ b/.github/workflows/sign-and-release-to-github.yml
@@ -64,17 +64,32 @@ jobs:
       - name: "Lint"
         run: npm run lint
 
-      - name: "Configure add-on for stage site"
+      - name: "Configure add-on for the stage site"
         run: npm run config:stage
+        if: github.ref_type == 'branch'
 
-      - name: "Set YYYY.MM.DD.minutes version"
+      - name: "Configure add-on for the production site"
+        run: npm run config:prod
+        if: github.ref_type == 'tag'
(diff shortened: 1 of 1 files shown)


.github/workflows/sign-and-release-to-github.yml
```diff
@@ -64,17 +64,32 @@ jobs:
...
         run: npm run lint
 
-      - name: "Configure add-on for stage site"
+      - name: "Configure add-on for the stage site"
         run: npm run config:stage
+        if: github.ref_type == 'branch'
 
-      - name: "Set YYYY.MM.DD.minutes version"
+      - name: "Configure add-on for the production site"
+        run: npm run config:stage
```
maxxcrawford, line 72: suggestion: I think this should be for prod. 

```suggestion
        run: npm run config:prod
```

maxxcrawford (changes requested):
Caught one issue - otherwise, this is good to go!

.github/workflows/sign-and-release-to-github.yml
Vinnl: Ooof good catch :sweat_smile: Also why I didn't update the docs yet: want to try it out properly with the next tag, because there's probably more mistakes. 073bda7

maxxcrawford (approved):
127.X 🛑

PR #384, opened 2022-07-19
Fix local labels disappearing when website updates

### LINKED ISSUE #383: Labels disappear from Relay Field Icon menu and dashboard when generating new mask from dashboard
**Build:**
Stage
Production

**Affected platforms:**
Windows 10
Android 12

**Browsers tested on:**
Firefox, latest release

**Preconditions:**
The Relay add-on is already installed.
The Data collection is turned OFF.
The VPN is set up in a country where Relay Premium is available.

**Steps to reproduce:**
1. Go to Relay page and log in using an existing account with or without a Premium subscription.
2. Go to a website that has an email input field.
3. Click the Relay Field Icon.
4. Click the Generate new mask button and click the Relay Field Icon again to observe the menu.
5. Go to Relay dashboard and click Generate new mask button.
6. Go back to the website, click the Relay Field Icon and observe the menu.

**Actual result:**
The mask labels disappear. The "From this website" and "All email masks" tabs are not available anymore. Only one list is displayed showing all the masks (website generated and dashboard generated). 

**Expected result:**
The mask labels are saved. The "From this website" and "All email masks" tabs are available and have the lists containing the correct masks.

**Notes:**
If after all the mentioned steps you go back to dashboard and Refresh the page -> the labels from dashboard disappear too.

labels not saved

Vinnl:
Fixes #383.

When the website informed the add-on that the list of masks was
updated (e.g. because the user generated a new mask on the
website), the add-on updates its local cache of masks, and tries to
copy over the labels it has stored locally for those masks.
However, the API to retrieve these local labels returns a Promise.
Since that wasn't `await`ed, the local labels were never found, and
then subsequently dropped them when replacing the local mask cache
with the updated mask list from the server.

(And we don't have TypeScript in the add-on that would've warned us about this :slightly_smiling_face: )

### DIFF
--- a/src/js/relay.firefox.com/get_profile_data.js
+++ b/src/js/relay.firefox.com/get_profile_data.js
@@ -44,7 +44,7 @@
     // Get the api token from the account profile page
     const profileMainElement = document.querySelector("#profile-main");
     const apiToken = profileMainElement.dataset.apiToken;
-    browser.storage.local.set({ apiToken });
+    await browser.storage.local.set({ apiToken });
 
     // API URL is ${RELAY_SITE_ORIGIN}/api/v1/
     const { relayApiSource } = await browser.storage.local.get("relayApiSource");
@@ -118,8 +118,8 @@
         : [];
       await browser.storage.local.set({
         relayAddresses:
-          applyLocalLabels(relayAddresses)
-          .concat(applyLocalLabels(domainAddresses)),
+          (await applyLocalLabels(relayAddresses))
+          .concat(await applyLocalLabels(domainAddresses)),
       });
     }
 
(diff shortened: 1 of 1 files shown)


src/js/relay.firefox.com/get_profile_data.js
```diff
@@ -44,7 +44,7 @@
     // Get the api token from the account profile page
     const profileMainElement = document.querySelector("#profile-main");
     const apiToken = profileMainElement.dataset.apiToken;
-    browser.storage.local.set({ apiToken });
+    await browser.storage.local.set({ apiToken });
```
Vinnl, line 47: Technically this change wasn't needed for this bugfix, but I thought it'd be cleaner to do it here too.

PR #389, opened 2022-07-20
Create releases when pushing tags

Vinnl:
The job was there (added in https://github.com/mozilla/fx-private-relay-add-on/pull/382), but it didn't run yet, and therefore the new tag did not build production artefacts yet.


codemist (approved):

codemist:
Going to test this out.

Vinnl:
Ah, I might have been too strict with the trigger format - it only runs on tags like `2.4.6`, not `2.4.6-chrome` :). Let's check this out next time we do a proper release then.

Issue #91, opened 2021-05-30
"If a site doesn't work, let us know"


ph818:
This link on the faq page (https://relay.firefox.com/faq) took me to the extension feedback page. I'm not sure if you want people to leave reviews of the extension with X website didn't work or not, or to click through to github and leave an issue here.

The website that (understandably) did not accept relay was https://cloud.ibm.com/registration

pendantry:
I'm in the same situation. Got a problem with Firefox Relay, followed that link inviting me to 'let you know' there's a problem, and... can't.

PR #390, opened 2022-07-25
Add Report Issue Panel

codemist:
This adds the Report issues feature on the add on where users can submit reports directly from the site where they are experiencing issues.

Inputs accepted:
- Accepts non-formatted URLs: eg. _test.com_, _test.co.uk_, _jane.app_, _www.jane.app_, 見.香港
- Disables submit button if the URL field is empty / does not contain a valid URL 

Demo:

https://user-images.githubusercontent.com/13066134/196756228-178ae784-69f1-48b8-bd8a-98db7dd13283.mov

Fixes: #91.


src/js/popup/popup.js
```diff
@@ -531,6 +530,86 @@ async function enableSettingsPanel() {
...
+    reportContent.classList.add("is-hidden");
+  });
+}
+
+async function reportURL() {
+  // Add Site URL placeholder
+  const currentPage = await getCurrentPage();
+  const url = new URL(currentPage.url);
+  const inputFieldUrl = document.querySelector('input[name="issue_on_domain"]');
+  inputFieldUrl.value = url.origin;
```
codemist, line 591: This ensures that the site is autofilled and is in the `https://` format, or backend cannot accept it.

codemist:
@say-yawn Added you as a reviewer just in case you wanted to spot check.

src/js/background/background.js
```diff
@@ -94,6 +94,40 @@ async function patchMaskInfo(method = "PATCH", id, data, opts=null) {
   return await response.json();
 }
 
+async function postReportWebcompatIssue(description) {
+  const { relayApiSource } = await browser.storage.local.get("relayApiSource");  
+  if (!relayApiSource) {
+    return;
+  }
+  const headers = await createNewHeadersObject({auth: true});
+  const reportWebCompatResponse = `${relayApiSource}/report_webcompat_issue`;
```
Vinnl, line 103: This variable name seems misleading - it contains a URL (I think?) rather than a response?

src/js/background/background.js
```diff
@@ -94,6 +94,40 @@ async function patchMaskInfo(method = "PATCH", id, data, opts=null) {
...
+    add_on_visual_issue: "",
+    email_not_received: "",
+    other_issue: "",
+  };
+
+  apiBody.issue_on_domain = description.issue_on_domain;
+  apiBody.email_mask_not_accepted = description.email_mask_not_accepted;
+  apiBody.add_on_visual_issue = description.add_on_visual_issue;
+  apiBody.email_not_received = description.email_not_received;
+  apiBody.other_issue = description.other_issue;
```
Vinnl, line 117: Heh, this works, but I'm just wondering why you didn't do this:

```suggestion
  const apiBody = {
    issue_on_domain: description.issue_on_domain,
    email_mask_not_accepted: description.email_mask_not_accepted,
    add_on_visual_issue: description.add_on_visual_issue,
    email_not_received: description.email_not_received,
    other_issue: description.other_issue,
  };
```

src/js/popup/popup.js
Vinnl: Do you mean that this ensures the protocol is part of the URL (which is OK), or that `http://` is rejected and only `https://` is accepted (which would be weirder)?

src/popup.html
```diff
@@ -1,170 +1,304 @@
 <!DOCTYPE html>
```
Vinnl, line 1: I did not actually review this file since the formatting changes make it hard to see what's actually new. If you could point out which parts you changed, I'd be happy to give it another look :)

Vinnl (approved):
For some reason I can't run the extension myself at the moment, so I've only looked at the code and not tested it myself, but assuming it works I see no blockers, with the caveat that I did not look at the HTML. And if the video recording you posted is anything to go by, it looks really good!

maxxcrawford (changes requested):
Getting an error when I try to report on a non-URL page (about: page, new tab). 

image


```
TypeError: Failed to construct 'URL': Invalid URL
    at reportURL (popup.js:589:15)
```

src/js/popup/popup.js
codemist: Sorry - both `http://` and `https://` are accepted.

src/js/popup/popup.js
```diff
@@ -531,6 +530,110 @@ async function enableSettingsPanel() {
...
+    }
+    // Remove from report if empty string
+    if (reportData[value] === "") {
+      delete reportData[value];
+    }
+  });
+  // Clean URL data to add "http://" before it if the url doesn't contain a protocol
+  if (!(reportData.issue_on_domain.startsWith("http://") || reportData.issue_on_domain.startsWith("https://"))) {
+    reportData.issue_on_domain = "http://" + reportData.issue_on_domain;
+  }
```
codemist, line 571: This allows the form to accept non-formatted URLs such as `www.test.com` or `test.com`.

src/js/popup/popup.js
```diff
@@ -531,6 +530,86 @@ async function enableSettingsPanel() {
...
+  // Add Site URL placeholder
+  const currentPage = await getCurrentPage();
+  const url = new URL(currentPage.url);
+  const inputFieldUrl = document.querySelector('input[name="issue_on_domain"]');
+  inputFieldUrl.value = url.origin;
+}
+
+ function showReportInputOtherTextField() {
+  const otherCheckbox = document.querySelector('input[name="issue-case-other"');
+  const otherTextField = document.querySelector('input[name="other_issue"');
```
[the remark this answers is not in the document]
codemist: Weirdly, it didn't.

src/js/popup/popup.js
```diff
@@ -531,6 +530,113 @@ async function enableSettingsPanel() {
...
+  reportIssueSubmitBtn.disabled = true;
+
+  // Allow for custom URL inputs
+  inputFieldUrl.addEventListener('input', () => {
+    reportIssueSubmitBtn.disabled = true;
+    // Ensure that the custom input is valid
+    if (isValidURL(inputFieldUrl.value)) {
+      reportIssueSubmitBtn.disabled = false;
+    }
+  });
```
codemist, line 610: This is what checks for the validity of custom inputted URLs.

maxxcrawford (commented):
Another top-level question/comment: 

- Has this been tested in Chrome?
- Is there any way for us to parse if the user reported from Chrome vs Firefox?

src/js/popup/popup.js
Vinnl: Instead of adding custom styling, you could also outsource form validation to the browser by calling [`.setCustomValidity`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setCustomValidity) (followed by `.reportValidity()`) on the URL input field, which should also prevent the form from being submitted. (You could also set `type="url" required` on the input field, I think, but that does require people to type `https://` themselves.)

Vinnl (commented):
Just calling out some things that struck me, but I'll leave the rest of the review to Maxx now :)

src/js/popup/popup.js
```diff
@@ -583,17 +585,42 @@ function showSuccessReportSubmission() {
   });
 }
 
-async function reportURL() {
+function isValidURL(string) {
+  // Disabled es-lint in the following line to accept regex
+  /* eslint-disable-next-line */
+  var expression = /(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g;
+  const res = string.match(expression);
+  return (res !== null)
```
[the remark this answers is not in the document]
codemist: Opted to check for if there is a dot inside of the inputted URL to account for non-english characters.

jwhitlock:
@maxxcrawford asked:

> Is there any way for us to parse if the user reported from Chrome vs Firefox?

I could see two ways:
1. The Add-on could report that it was from the Chrome store or Firefox Add-ons. Is this information available in the add-on?
2. The backend could look at the `User-Agent` header. Looking at `/metrics-event`, we get the user agent, but we'll need to do additional parsing (and probably bring in a new library) to turn this into "Chrome", "Firefox", and "Something else"

Vinnl:
The add-on has access to [`runtime.getBrowserInfo`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getBrowserInfo) and [`runtime.getPlatformInfo`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getPlatformInfo), but I think the backend will need an update to accept that data.

codemist:
@jwhitlock Se Yeon mentioned that ideally we should collect that data on the frontend and pass it.

src/popup.html
```diff
@@ -1,170 +1,304 @@
...
+  <script src="js/shared/metrics.js"></script>
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+</head>
 
-      <sign-up-panel class="sign-up-panel hidden">
-        <span class="welcome i18n-content" data-i18n-message-id="popupSignUpPanelWelcome"></span>
-        <p class="intro i18n-content" data-i18n-message-id="popupSignUpPanelIntro"></p>
-        <a class="button sign-in-btn blue-primary-btn close-popup-after-click login-link i18n-content" data-event-action="click" data-event-label="panel-continue-btn" data-i18n-message-id="popupSignUpPanelContinue" href=""></a>
-      </sign-up-panel>
+<body class="fx-relay-panel fx-relay">
```
maxxcrawford, line 19: issue (blocking): Following up on @Vinnl comment about this diff being big. I was able to hide whitespace on this review and noticed you have nested `<body><head>` tags below. Seems like maybe a bad rebase? 

Can you review / reset this file and contain your changes to `<report-issue>`

src/popup.html
```diff
@@ -1,170 +1,304 @@
...
+                data-event-label="panel-leave-feedback-link" data-i18n-message-id="popupSettingsLeaveFeedback"></a>
+              <a href="https://www.mozilla.org/privacy/firefox-relay/?utm_source=fx-relay-addon&utm_medium=popup"
+                class="setting-label setting-link close-popup-after-click i18n-content" data-event-action="click"
+                data-event-label="panel-privacy-link" data-i18n-message-id="popupSettingsPrivacy"></a>
+              <a href="https://www.mozilla.org/about/legal/terms/firefox-relay/?utm_source=fx-relay-addon&utm_medium=popup"
+                class="setting-label setting-link close-popup-after-click i18n-content" data-event-action="click"
+                data-event-label="panel-terms-of-service" data-i18n-message-id="popupSettingsTermsOfService"></a>
+              <!-- Report Issue Panel Prompt -->
+              <a class="setting-label setting-link settings-report-issue i18n-content" data-event-action="click"
+                data-event-label="panel-report-issue" data-i18n-message-id="popupSettingsReportIssue"></a>
```
maxxcrawford, line 83: continued from previous comment above: this snippet should also be included on the file reset.

maxxcrawford (changes requested):
Great work! Thanks for rolling with the scope change on browser telemetry. 

I found a blocking issue with the `src/popup.html` file. See below. Happy to walk you through the issues directly. LMK!

src/js/popup/popup.js
```diff
@@ -555,6 +553,110 @@ async function enableSettingsPanel() {
...
+      const eventLabel = document.body.classList.contains("show-report-issue") ? "opened-report-issue" : "closed-report-issue";
+      if (document.body.classList.contains("show-report-issue")) {
+        sendRelayEvent("Panel", "click", eventLabel);
+      }
+    });
+  });
+  const reportForm = document.querySelector(".report-issue-content");
+  setURLwithIssue();
+  showReportInputOtherTextField();
+  showSuccessReportSubmission();
```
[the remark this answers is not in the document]
codemist: Saving this. This is a much cleaner pattern, I'll definitely integrate this method into larger scoped features

src/popup.html
codemist: Oh wow, thank you for pointing that out. I didn't even notice any visible changes to the panel with this error which is weird.

codemist:
@maxxcrawford Thanks for catching the errors in `popup.html`, that was a complete oversight. The differences now should be more clear, I reformatted the entire file with prettier and I thought that was what was causing the huge change and thought nothing of it. It should not have any bad rebase artifacts now.

src/popup.html
maxxcrawford: @Vinnl This comment was what made me a closer look on my review. Thanks for making this note!

src/popup.html
```diff
@@ -123,46 +206,61 @@ <h1 class="onboarding-h1"></h1>
...
-                <button class="premium-panel-nav next-panel i18n-attribute" data-direction="1" data-i18n-attribute="aria-label" data-i18n-message-id="popupOnboardingShowNext" aria-label=""></button>
+                <button class="premium-panel-nav next-panel i18n-attribute" data-direction="1"
+                  data-i18n-attribute="aria-label" data-i18n-message-id="popupOnboardingShowNext"
+                  aria-label=""></button>
               </div>
-              
+
             </div>
           </div>
+          </div>
```
maxxcrawford, line 228: suggestion (non-blocking): Remove extra `<div>` tag.  
```suggestion
```

src/popup.html
```diff
@@ -19,47 +19,124 @@
...
+              <li>
+                <!-- Case Four -->
+                <input type="checkbox" id="issue-case-other" name="issue-case-other" />
+                <label class="i18n-content" for="issue-case-other"
+                  data-i18n-message-id="popupReportIssueCaseOther"></label>
+              </li>
+              <li>
+                <!-- 'Other' option input field -->
+                <input class="report-other-field is-hidden" type="text" name="other_issue"
+                  id="other_issue" name="other_issue" />
```
maxxcrawford, line 107: suggestion (non-blocking): Duplicate `name` attribute. 
```suggestion
                <input class="report-other-field is-hidden" type="text" name="other_issue"
                  id="other_issue" />
```

maxxcrawford (approved):
@codemist Thanks for the quick uppdate to the `src/popup.html` file! I found two outstanding issues in the HTML. ([Thanks, W3C Validator](https://validator.w3.org)) Fix those and we can ship. (Please double-back and review all still open / non-resolved comments from @Vinnl and myself, in case there was other follow-up needed.) 

Additionally, there's still lots of whitespace changes to `src/popup.html` that I would love to not be there, but if they stick, that's okay. 😬

src/popup.html
```diff
@@ -19,47 +19,124 @@
...
+            <label class="report-label i18n-content" for="issue_on_domain" data-i18n-attribute="title"
+              data-i18n-message-id="popupReportIssueSiteLabelHeadline"></label>
+            <input type="text" id="issue_on_domain" name="issue_on_domain" value="" />
+          </div>
+          <div class="report-section">
+            <!-- Select issue checkboxes -->
+            <label class="report-label i18n-content" data-i18n-message-id="popupReportIssueSituationHeadline"></label>
+            <ul>
+              <li>
+                <!-- Case One -->
```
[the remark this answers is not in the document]
codemist: Gotcha, addressed all of the comments. I'll fix the whitespace differences.

PR #393, opened 2022-08-30
Add Countdown Timer on Addon

codemist:
Preview: 
image

Features:
- Adds the countdown timer (to be submitted to AMO + Chrome on the **6th of September**)
- Adds email trackers removed data to the dashboard statistics (available to both premium & non-premium)

**How to test:**
With a non-premium account, check that the popup panel displays the countdown timer instead of the tips view. 
With a premium account, check that the countdown timer is not visible to users. 
Set the date on your calendar to after the 27th of September, and then check that the non-premium view resets back to its default state.


codemist:
For some reason, I can't get the font `Metropolis` to work. I'll open another PR separately for this, but seeing as it already took me several hours to debug, I don't want this work to get blocked.

src/css/popup.css
```diff
@@ -479,17 +494,21 @@ onboarding-panel {
 .email-domain-illustration {
   margin: 0 auto;
   display: block;
+  height: 80px;
```
codemist, line 497: Sets a fixed height for the image to make sure all of the UI components fit (including the new email trackers removed data), since there isn't a lot of real estate on the popup panel.

src/popup.html
```diff
@@ -97,9 +150,13 @@ <h1 class="onboarding-h1"></h1>
...
                   <div class="dashboard-stats emails-forwarded"></div>
                 </li>
+                <li class="dashboard-info">
+                  <span class="email-trackers-removed-text i18n-content" data-i18n-message-id="popupEmailTrackersRemoved"></span>
+                  <div class="dashboard-stats email-trackers-removed"></div>
+                </li>
               </ul>
             </div>
-            <!-- Premium Dashboard-->
+            <!-- Register Domain Component -->
```
Vinnl, line 159: Good work also cleaning up outdated comments. (Have I gone on one of my rants about default syntax highlighting themes always de-emphasising comments yet? :sweat_smile:)

src/js/popup/popup.js
```diff
@@ -53,6 +53,80 @@ function getEducationalStrings() {
...
+      remainingDays,
+      remainingHours,
+      remainingMinutes,
+      remainingSeconds,
+    }
+  }
+
+    function setRemainingTimeParts() {
+      const countdownTimer = document.querySelector(".countdown-timer");
+      const introPricingOfferEndDate = new Date(Date.UTC(2022, 8, 27, 16));
```
Vinnl, line 111: To be safe, it might be better to obtain this date from the API (it's returned from `/runtime_data/` now). Just in case we need to change the date for whatever reason, it'd be great not to have to scramble to get an updated version of the add-on published.

Vinnl: Actually, I see that in `background.js`, we're still getting per-country Premium availability from the deprecated `/premium_countries` API, rather than `/runtime_data`. So this could probably be relatively easily done by renaming `storePremiumAvailabilityInCountry` to `storeRuntimeData`, make it point to the `/runtime_data` endpoint, and then in addition to `premiumCountries` to also store `INTRO_PRICING_END`.

Vinnl (approved):
Work as expected. Only small nit is that the countdown timer shortly shows up when I change my system time beyond the deadline. Probably not a blocker (?), but if there's an easy fix that might be worth it.

https://user-images.githubusercontent.com/4251/187888686-61cc77e9-0153-455a-8bae-f0853a205d56.mp4

As for the Metropolis issue, I'm not quite sure what would cause that, but it might have to do with whatever Maxx created `setCustomFonts` for in `/src/js/shared/utils.js`, so you could take a look at that too. I don't expect it to be worth it to spend too much time on it though - folks will survive if it's shown in a different font :grin: 

I also added a comment suggesting getting the end date from the API, but that could also be done in a separate PR (or not at all, if we're confident about this date or about our ability to update the add-on in time if needed).

src/popup.html
codemist: Thanks :) Also, could you slack me the behaviour you're talking about? Github's been having issues displaying videos.
image

src/js/popup/popup.js
codemist: Done, the end date and premium countries is being fetched from `/runtime_data` now.

src/js/popup/popup.js
```diff
@@ -342,11 +432,11 @@ async function showRelayPanel(tipPanelToShow) {
...
-  //Dashboard Data
-  const { aliasesUsedVal } = await browser.storage.local.get("aliasesUsedVal");
-  const { emailsForwardedVal } = await browser.storage.local.get("emailsForwardedVal");
-  const { emailsBlockedVal } = await browser.storage.local.get("emailsBlockedVal");
-
+  // Set End Date here
+  const introPricingEndDateISO = (await browser.storage.local.get("introPricingEndDate"))?.introPricingEndDate.INTRO_PRICING_END;
+  const introPricingOfferEndDate = new Date(introPricingEndDateISO);
+  
+  showCountdownTimer(introPricingOfferEndDate);
```
codemist, line 439: End date is set here.

groovecoder (changes requested):

src/js/background/background.js
```diff
@@ -87,10 +87,10 @@ async function patchMaskInfo(method = "PATCH", id, data, opts=null) {
...
 
-async function storePremiumAvailabilityInCountry() {
+async function storeRuntimeData() {
   // If we already fetched Premium availability in the past seven days,
   // don't fetch it again.
-  const existingPremiumAvailability = (await browser.storage.local.get("premiumCountries"))?.premiumCountries;
+  const existingPremiumAvailability = (await browser.storage.local.get("premiumCountries")).premiumCountries?.PREMIUM_PLANS;
   if (typeof existingPremiumAvailability === "object" && existingPremiumAvailability.fetchedAt > (Date.now() - 7 * 24 * 60 * 60 * 1000)) {
     return;
   }
```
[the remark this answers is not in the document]
Vinnl: Hmm yeah, I'd expect that to happen. If the `introPricingEndDate` is not found, [`new Date(undefined)` is called](https://github.com/mozilla/fx-private-relay-add-on/blob/49aab832bf386a754ee25da0c2c14a62ee202e59/src/js/popup/popup.js#L441-L442), which results in an invalid Date object, for which `.getTime()` returns `NaN`, which I'd expect to cause all kinds of problems.

I think to be sure we'd probably best disregard whatever value is already stored for `premiumCountries` in local storage (possibly even just deleting any existing values, if present), and then instead checking for a new value `runtimeData` instead?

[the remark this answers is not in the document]
groovecoder: > Edit: though of course, that might be problematic for other references to `premiumCountries` in the code. Ugh, caching and no types... OK, then alternatively instead of just checking for `existingPremiumAvailability.fetchedAt`, we check whether either that _or_ existingIntroPricingEndDate.fetchedAt` is nonexistent or more than seven days ago?

Yeah, I can't believe I find myself WANTING types now. 😜 

I agree - we should probably check for both `existingPremiumAvailability` AND ` existingIntroPricingEndDate` - and if either of them doesn't exist or is 7+ day sold, re-fetch.

src/js/background/background.js
```diff
@@ -87,11 +87,21 @@ async function patchMaskInfo(method = "PATCH", id, data, opts=null) {
...
+  // don't fetch it again.
+  const checkingRemainingDays = Date.now() - 7 * 24 * 60 * 60 * 1000;
+  if (typeof existingPremiumAvailability === "object" && 
+      typeof existingIntroPricingEndDate === "object" && 
+      existingPremiumAvailability.fetchedAt > checkingRemainingDays &&
+      existingIntroPricingEndDate.fetchedAt > checkingRemainingDays
+      ) 
+      {
     return;
   }
```
codemist, line 106: Added this in as per Luke's suggestion:

> check for both existingPremiumAvailability AND existingIntroPricingEndDate - and if either of them doesn't exist or is 7+ day sold, re-fetch

@Vinnl @groovecoder Could you test this again and make sure it still works?

Vinnl: Just tested it by installing [this stage add-on](https://github.com/mozilla/fx-private-relay-add-on/actions/runs/2975980121), signing in, then installing [this one](https://github.com/mozilla/fx-private-relay-add-on/actions/runs/2979920257), and it appeared to work well.

groovecoder (approved):
Thanks for being extra thorough on it. This should help prevent us from having to make a 2.4.7 to clean this up post-launch.

PR #399, opened 2022-09-21
Pull in updates from the l10n repo on a schedule

Vinnl:


flodolo:
Review ping 🙏🏼

.github/workflows/l10n-sync.yml
```diff
@@ -0,0 +1,28 @@
+name: Fetch latest strings from the l10n repo
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
```
groovecoder, line 5: can we add a comment saying what this schedule is? I.e., `every night at midnight UTC` ?

groovecoder (changes requested):

.github/workflows/l10n-sync.yml
Vinnl: Good point: d76ac61

groovecoder (approved):

PR #398, opened 2022-09-15
Add Phone masking news announcement

codemist:
**OCT 4 Update: Bundle announcements have been removed. If the phones flag is on, only the phone announcement should be visible to premium & non-premium users in phone-masking available countries.**

**Cases:**

**1. Non-premium user**

- [x] Condition: bundle & phone flags are on, and user is in US/CAN 
> image
- [x] Condition: either bundle/phone flag is off and/or user is not in US/CAN (in other words, if any of these conditions aren’t met, do not show promotional announcements)
> image

- [x] Condition: max aliases reached and user is not in US/CAN and/or bundle/phone flag is off (in other words, continue to show the promotional panel even when there are no remaining masks left if all of the conditions (bundle/phone flag, user is in US/CAN) are met)

> image

**2. Relay Premium User:**
- [x] Condition: bundle & phone flags are on, and user is in US/CAN (all conditions must be met) 

> image

- [x] Condition: either bundle/phone flag is off and/or user is not in US/CAN (in other words, if any of these conditions aren’t met, do not show promotional announcements)

> image

3. On Chrome
- [x] The same behaviour, but without the ‘Sign back in’ announcement.

4. CTA:
- [x] Make sure the upgrade CTAs take the user to the `/premium#pricing` section on the home page.

PR #401, opened 2022-10-04
Remove references to old Premium plans

Vinnl:
This removes references to the old `PREMIUM_PLANS`, which will be removed from the API, instead looking at `PERIODICAL_PREMIUM_PLANS` to determine whether Premium is available for a user. (The new object has new plan IDs, grouped by monthly and yearly subscriptions, rather than only the monthly subscription being available.)

I set this PR to target #398, because that already added the new plans to local storage - this also removes the old plan.

### DIFF
--- a/src/js/background/background.js
+++ b/src/js/background/background.js
@@ -95,7 +95,7 @@ async function patchMaskInfo(method = "PATCH", id, data, opts=null) {
 }
 
 async function storeRuntimeData(opts={forceUpdate: false}) {  
-  const existingPremiumAvailability = (await browser.storage.local.get("premiumCountries")).premiumCountries;
+  const existingPremiumAvailability = (await browser.storage.local.get("periodicalPremiumPlans")).periodicalPremiumPlans;
   // If we already fetched Premium availability in the past seven days,
   // don't fetch it again.
   const checkingRemainingDays = Date.now() - 7 * 24 * 60 * 60 * 1000;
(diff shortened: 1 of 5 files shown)


maxxcrawford:
What is the best way to test/verify this? Which flags should I enable?

Vinnl:
@maxxcrawford No flag changes should be needed; there just should be no regressions, especially to all Premium-related features. And I guess for extra thoroughness, you could check out [this branch](https://github.com/mozilla/fx-private-relay/pull/2602) for the website - that should break with the current add-on, but not with the code in this branch.

maxxcrawford:
@Vinnl Awesome. I'll do that. Thanks!

src/js/background/background.js
```diff
@@ -90,16 +90,13 @@ async function patchMaskInfo(method = "PATCH", id, data, opts=null) {
...
+  const existingPremiumAvailability = (await browser.storage.local.get("periodicalPremiumPlans")).periodicalPremiumPlans;
   
   // If we already fetched Premium availability in the past seven days,
   // don't fetch it again.
   const checkingRemainingDays = Date.now() - 7 * 24 * 60 * 60 * 1000;
   if (typeof existingPremiumAvailability === "object" && 
-      typeof existingIntroPricingEndDate === "object" && 
-      existingPremiumAvailability.fetchedAt > checkingRemainingDays &&
-      existingIntroPricingEndDate.fetchedAt > checkingRemainingDays
+      existingPremiumAvailability.fetchedAt > checkingRemainingDays
```
maxxcrawford, line 99: This may have a conflict now bc #414 landed (my bad)

maxxcrawford (approved):
Good to go (except for possible conflict I called out)

src/js/background/background.js
Vinnl: No worries, there was a bit of overlap too - rebased.

PR #404, opened 2022-10-06
Fix #403 - Added error catching to getCurrentPageHostname() bg message function to ensure settings options in the pop-up work on internal browser pages

### LINKED ISSUE #403: If on a non-URL page (about: pages), the "Show Relay icon" setting cannot be toggled off/on

maxxcrawford:
Fixed #403 

While testing the add-on on another PR, I ran across this edge case. 

To test: 
- Open add-on in either desktop browser
- While on the entry/add-on login page, click the pop-up panel icon
- Go to settings panel
- Click first setting "Show Relay icon in email fields on websites" 
- **Expected:** The setting should toggle off (or back on if off previously) as expected
- Go to a login page and confirm setting works as expected (facebook.com, accounts.spotify.com, etc)

### DIFF
--- a/src/js/background/background.js
+++ b/src/js/background/background.js
@@ -361,7 +361,11 @@ async function displayBrowserActionBadge() {
 browser.runtime.onMessage.addListener(async (m, sender, _sendResponse) => {
   let response;
   const currentPage = await getCurrentPage();
-  const url = new URL(currentPage.url);
+  let url = false;
+
+  if (currentPage.url) {
+    url = new URL(currentPage.url);
+  }
 
   switch (m.method) {
     case "displayBrowserActionBadge":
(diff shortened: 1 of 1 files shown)


src/js/background/background.js
```diff
@@ -361,7 +361,11 @@ async function displayBrowserActionBadge() {
 browser.runtime.onMessage.addListener(async (m, sender, _sendResponse) => {
   let response;
   const currentPage = await getCurrentPage();
-  const url = new URL(currentPage.url);
+  let url = false;
```
maxxcrawford, line 364: LMK if there's a better way to handle this error than using `let` and resetting inside an `if` statement.

Vinnl: Better is subjective, but an _alternative_ way is

```suggestion
  const url = typeof currentPage.url === "string" ? new URL(currentPage.url) : false;
```

_Another_ alternative way is to remove this line altogether, and below do:

```diff
- if (url) { response = url.hostname }
+ if (url) { response = (new URL(currentPage.url))l.hostname }
```

But they're all fine by me :smile:

Vinnl (approved):

src/js/background/background.js
maxxcrawford: Solid. I was able to do a mix of the two. Much cleaner! 

_(and comment added)_

PR #407, opened 2022-10-06
Don't generate .xpi's when pushing a tag

Vinnl:
Generating an XPI will push it to addons.mozilla.org, which will then no longer accept new uploads of that version. Since the .xpi we upload is unlisted, we can't use it as our release.

Instead, just like for Chrome, we'll need to upload the .zip file to AMO.


.github/workflows/sign-and-release-to-github.yml
```diff
@@ -115,7 +116,7 @@ jobs:
 
       - uses: actions/upload-artifact@v2
         with:
-          name: fx-private-relay-stage-chromium.zip
+          name: fx-private-relay-extension.zip
```
Vinnl, line 119: I renamed this since AFAIK this same zip can be uploaded to both the Chrome web store and AMO.

maxxcrawford (approved):
Thanks for the quick fix!

PR #409, opened 2022-10-08
Update README to describe CI-based release process

Vinnl:
This should hopefully avoid issues with misaligned local config changes (e.g. the add-on still pointing to the stage environment).


maxxcrawford (approved):
Perfect!

PR #411, opened 2022-10-10
Update CI to remove the .github submodule folder during sign and release

maxxcrawford:
This PR adds a `.github` folder removal step from the /src folder during the automation "Sign and Release" step.


package.json
```diff
@@ -22,7 +22,8 @@
     "package:chrome": "zip -r web-ext-artifacts/chrome_relay-${npm_package_version}.zip ./src",
     "test": "echo \"Error: no test specified\" && exit 1",
     "restore-locales-github": "cd src/_locales && git restore .github/workflows/update-upstream-relay-repo.yml",
-    "web-ext-run": "rm -rf src/_locales/.github && web-ext run -s src/",
+    "remove-locales-github": "rm -r src/_locales/.github",
+    "web-ext-run": "rm -r src/_locales/.github && web-ext run -s src/",
```
Vinnl, line 26: Oh, we remove it in _our_ scripts! I thought `web-ext` did that :sweat_smile: In that case, would something like this make sense?

```suggestion
    "web-ext-run": "npm run remove-locales-github && web-ext run -s src/ && npm run restore-locales-github",
```

Vinnl: Though I guess if you terminate the script with <kbd>Ctrl</kbd>+<kbd>C</kbd> this doesn't do anything. But might still be useful if you just close the browser?

Vinnl (approved):
Looks good! Left suggestion that's probably nonsense, so feel free to ignore :)

package.json
maxxcrawford: Bleh. Current form is not ideal. I punted to an issue for now. https://github.com/mozilla/fx-private-relay-add-on/issues/412

PR #415, opened 2022-10-13
Fix Phone Masking Announcement UI Bugs

codemist:
This fixes the pagination error on the add on. It should now be displaying the correct number of panels on a premium value for a user not subscribed to phones / is not located in US/CAN.

This PR fixes two tickets -

MPP-2461 (Show the correct number of panels in the pagination bar):
image
How to test:
- With a premium account on a phone plan or a premium account outside of US/CAN, show 4 total panels in the pagination bar.

MPP-2071 (Do not show phone masking announcement to users who have purchased phones)
How to test:
- With an account that has purchased phones, ensure that the add on reverts back to the original state.

### DIFF
--- a/src/js/popup/popup.js
+++ b/src/js/popup/popup.js
@@ -347,9 +347,7 @@ async function showRelayPanel(tipPanelToShow) {
     if (panelStrings.longText) {
       educationBodyEl.classList.add("small-font-size");
     }
-
-    const totalPanels = Object.keys(premiumPanelStrings.announcements).length;
-    setPagination(panelId, totalPanels);
+    setPagination(panelId);
 
     educationHeadlineEl.textContent = panelStrings.tipHeadline;
     educationBodyEl.textContent = panelStrings.tipBody;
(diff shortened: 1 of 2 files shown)


src/js/popup/popup.js
```diff
@@ -442,8 +446,8 @@ async function showRelayPanel(tipPanelToShow) {
   remainingAliasMessage.textContent = browser.i18n.getMessage("popupRemainingAliases_2_mask", [numRemaining, maxNumAliases]);
   const getUnlimitedAliases = document.querySelector(".premium-cta");
   getUnlimitedAliases.textContent = browser.i18n.getMessage("popupGetUnlimitedAliases_mask");
-
   document.body.classList.add("relay-panel");
+
```
maxxcrawford, line 450: nit: white space 
```suggestion
```

src/js/popup/popup.js
```diff
@@ -272,12 +272,15 @@ async function showRelayPanel(tipPanelToShow) {
   // const isBundleAvailableInCountry = (await browser.storage.local.get("bundlePlans")).bundlePlans.BUNDLE_PLANS.available_in_country;
   const isPhoneAvailableInCountry = (await browser.storage.local.get("phonePlans")).phonePlans.PHONE_PLANS.available_in_country;
   
-  const phoneMaskingAvailable =    await checkWaffleFlag("phones") && isPhoneAvailableInCountry;
+  // If user has a phone plan, don't show the phone masking promo
+  const hasPhone = (await browser.storage.local.get("has_phone")).has_phone;
```
maxxcrawford, line 276: Question: Are there any additional data catches/failsafes if this check doesn't happen? _(Re: what happened last release)_

maxxcrawford (approved):
One nit and one question, but otherwise verified locally. Good to go!

src/js/popup/popup.js
codemist: Shouldn't the earlier fix ensure that profile data becomes updated upon a version release?

codemist: I also tested this with a new Nightly profile and setting up `xpinstall.signatures.required` to false etc. It _should_ just revert back to the original state if the fetch isn't successful. Let's make sure QA touches this first before submitting too.

PR #422, opened 2022-10-25
MPP-2207 - Enable bundle announcement

codemist:
This enables the bundle announcement that was omitted from the last release, as well as introduces some state management to check which of the two primary promos (or both) phone masking and bundle to surface depending on the user's profile data.

Preview in [this ticket](https://mozilla-hub.atlassian.net/jira/software/c/projects/MPP/boards/198?modal=detail&selectedIssue=MPP-2207). 
### **Test cases:**

_Note that for free/premium users, behaviour should be the same._

**Phone announcement requirements:** US/CAN user, phone flag on, phone plan not yet purchased
**Bundle announcement requirements:**  US/CAN user, bundle flag on, bundle plan not purchased

**Show both phone + bundle announcements:**
- For a user in US/CAN, with the phone + bundle flag on, without a phone or bundle plan purchased, show both phone masking + bundle announcement

**Only phone announcement visible:**
- For a user with phone announcement requirements met
- For a user who doesn't meet at least one of the criteria for bundle excluding US/CAN, so bundle flag off / bundle plan purchased

**Only bundle announcement visible:**
- For a user with bundle announcement requirements met
- For a user who doesn't meet at least one of the criteria for phone excluding US/CAN, so phone flag off / phone plan purchased

**Hide both phone + bundle announcements, revert to original view:**
- For a user who doesn't meet at least one of the criteria for both phone and bundle announcement requirements. Eg. phone and bundle plans both purchased / non-US/CAN user.

### DIFF
--- a/src/js/popup/popup.js
+++ b/src/js/popup/popup.js
@@ -9,9 +9,8 @@ async function checkWaffleFlag(flag) {
   return false;
 }
 
-
 async function getPromoPanels() {
-  const savings = "40%"; // For "Save 50%!" in the Bundle promo body
+  const savings = "40%"; // For "Save 40%!" in the Bundle promo body
   const getBundlePlans = (await browser.storage.local.get("bundlePlans")).bundlePlans.BUNDLE_PLANS;
   const getBundlePrice = getBundlePlans.plan_country_lang_mapping[getBundlePlans.country_code].en.yearly.price;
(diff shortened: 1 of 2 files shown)


Vinnl (approved):
I did not run it myself (I think we might need a bump of the `web-ext` version; I think the current version had a bug that kept trying to run it under MV3 after you've done that once, and that that is the reason I can't run the add-on on my machine?), but nothing jumps out to me in the code :)

PR #416, opened 2022-10-13
DNM/WIP: Manifest v3 Migration (Firefox Preview) 

maxxcrawford:
- WIP - Manifest v3 Migration
- Update build command to use Firefox Preview

PR #426, opened 2022-10-31
Move all fetch() requests to background scripts

maxxcrawford:
### Summary: 
This is part of a pre-migration to Manifest v3. The only content script where `fetch()` requests were happening was in the `get_profile_data.js` script, which [only runs on the `/accounts/profile/*` pages](https://github.com/mozilla/fx-private-relay-add-on/blob/main/src/manifest.json#L55-L60). 

**Links**: 
- See #416 for more context
- Related: [MPP-2505](https://mozilla-hub.atlassian.net/browse/MPP-2505)

### Testing:

* Run on add-on, log in/sign up for Relay account
* Expected: All data for the add-on is gathered as normal from the homepage without any breakage

⚠️  Be sure to test with both free / premium accounts with privacy/label storage toggled on and off. (4 possible test scenarios)

### DIFF
--- a/src/js/background/background.js
+++ b/src/js/background/background.js
@@ -25,6 +25,31 @@ browser.runtime.onInstalled.addListener(async (details) => {
   }
 });
 
+async function fetchApiRequest(url, fetchMethod = "GET", body = null, opts=null) {
+  const headers = new Headers();
+
+  const { csrfCookieValue } = await browser.storage.local.get("csrfCookieValue");
+  
+  headers.set("X-CSRFToken", csrfCookieValue);
+  headers.set("Content-Type", "application/json");
+  headers.set("Accept", "application/json");
+
+  if (opts && opts.auth) {
+    const apiToken = await browser.storage.local.get("apiToken");
+    headers.set("Authorization", `Token ${apiToken.apiToken}`);
+  }
+
+  const response = await fetch(url, {
+    mode: "same-origin",
+    method: fetchMethod,
+    headers: headers,
+    body,
+  });
+
+  const answer = await response.json();
+  return answer;
+}
+
 // This function is defined as global in the ESLint config _because_ it is created here:
 // eslint-disable-next-line no-redeclare, no-unused-vars
 async function getAliasesFromServer(method = "GET", opts=null) {
@@ -366,6 +391,14 @@ browser.runtime.onMessage.addListener(async (m, sender, _sendResponse) => {
       break;
     case "iframeCloseRelayInPageMenu":
       browser.tabs.sendMessage(sender.tab.id, {message: "iframeCloseRelayInPageMenu"});
+      break;
+    case "fetchApiRequest": 
+      response = await fetchApiRequest(
+        m.url, 
+        m.fetchMethod,
+        m.body,
+        m.otps
+      );
       break;
     case "fillInputWithAlias":
       browser.tabs.sendMessage(sender.tab.id, m.message);
--- a/src/js/relay.firefox.com/get_profile_data.js
+++ b/src/js/relay.firefox.com/get_profile_data.js
@@ -46,6 +46,24 @@
     const apiToken = profileMainElement.dataset.apiToken;
     await browser.storage.local.set({ apiToken });
 
+    const cookieString =
+    typeof document.cookie === "string" ? document.cookie : "";
+    
+    const cookieStringArray = cookieString
+      .split(";")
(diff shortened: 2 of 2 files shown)


src/js/background/background.js
```diff
@@ -367,6 +392,14 @@ browser.runtime.onMessage.addListener(async (m, sender, _sendResponse) => {
     case "iframeCloseRelayInPageMenu":
       browser.tabs.sendMessage(sender.tab.id, {message: "iframeCloseRelayInPageMenu"});
       break;
+    case "fetchApiRequest": 
+      response = await fetchApiRequest(
+        m.url, 
+        m.fetchMethod,
+        m.body,
+        m.otps
```
maxxcrawford, line 400: Question: I didn't run into any issues but is there any danger passing null/undefined values here? (Most of the time when this message is called, `fetchMethod`, `body` and `opts` are blank.

src/js/background/background.js
```diff
@@ -367,6 +393,14 @@ browser.runtime.onMessage.addListener(async (m, sender, _sendResponse) => {
     case "iframeCloseRelayInPageMenu":
       browser.tabs.sendMessage(sender.tab.id, {message: "iframeCloseRelayInPageMenu"});
       break;
+    case "fetchRequestFromBackground": 
+      response = await fetchRequestFromBackground(
+        m.url, 
+        m.fetchMethod,
+        m.body,
+        m.opts
```
maxxcrawford, line 401: Question: I didn't run into any issues but is there any danger passing null/undefined values here? (Most of the time when this message is called, `fetchMethod`, `body` and `opts` are blank.

src/js/relay.firefox.com/get_profile_data.js
```diff
@@ -46,57 +46,36 @@
...
+        cookieValue.trim(),
+      ]);
+
+    const [_csrfCookieKey, csrfCookieValue] = cookieStringArray.find(
+      ([cookieKey, _cookieValue]) => cookieKey === "csrftoken"
+    );
+
+    await browser.storage.local.set({ csrfCookieValue: csrfCookieValue });
+
+
```
Vinnl, line 66: I don't think you need a CSRF cookie if you're not making the request from the Relay website?

```suggestion
```

src/js/background/background.js
```diff
@@ -25,6 +25,32 @@ browser.runtime.onInstalled.addListener(async (details) => {
   }
 });
 
+
+async function fetchRequestFromBackground(url, fetchMethod = "GET", body = null, opts=null) {
```
Vinnl, line 29: Wouldn't it make sense to either move `fetchMethod` onto `opts`, or alternatively, to just accept the entire `init` parameter (i.e. the second parameter to `fetch`)?

src/js/background/background.js
Vinnl: I wouldn't think so, but your asking the question makes me nervous :P What danger were you thinking about?

Vinnl (approved):
Unexpected/error responses from the API will probably get even more opaque, but I think we're already not really handling them, so as far as I'm concerned that's not really a regression :)

src/js/relay.firefox.com/get_profile_data.js
maxxcrawford: I dug into the git-blame on the cookie string origins. It was me. 🪞👨 

https://github.com/mozilla/fx-private-relay-add-on/commit/200e2fbe6236fc5b2c32295cd7a62e6534e1750e

I think the initial reasoning was to verify if a user had the server-storage pref enabled? I'll remove this and confirm this doesn't break anything.

src/js/background/background.js
maxxcrawford: Rewrote this function so that if the URL param is not included, it fails. Everything else has a defined "default". 

https://github.com/mozilla/fx-private-relay-add-on/blob/mv3-spike-followup-fetch-migration/src/js/background/background.js#L40-L44

src/js/relay.firefox.com/get_profile_data.js
maxxcrawford: See 25255d1bed74c5164629f9c406df84270edfb21d

Vinnl: Ah OK, I wasn't aware that there was already code in `background.js` that read `csrfCookieValue` from the extension storage. AFAIK that's not needed, but it's outside of the scope of this PR then anyway.

Vinnl (approved):

PR #433, opened 2022-11-11
Add a "Use phone number mask" context menu option

Vinnl:
String: ~https://github.com/mozilla-l10n/fx-private-relay-add-on-l10n/pull/42~

### DIFF
--- a/src/js/background/context-menu.js
+++ b/src/js/background/context-menu.js
@@ -30,6 +30,13 @@ const staticMenuData = {
     visible: true,
     contexts: ["all"],
   },
+  insertPhoneMask: {
+    id: "fx-private-relay-insert-phone-mask",
+    title: browser.i18n.getMessage("pageInputIconInsertPhoneMask"),
+    enabled: true,
+    visible: true,
+    contexts: ["all"],
+  },
   upgradeToPremium: {
     id: "fx-private-relay-get-unlimited-aliases",
     title: browser.i18n.getMessage("pageInputIconGetUnlimitedAliases_mask"),
@@ -163,6 +170,11 @@ const relayContextMenus = {
           aliases
         );
       }
+    }
+
+    const phoneMask = await relayContextMenus.utils.getPhoneMask();
+    if (phoneMask) {
+      await relayContextMenus.menus.create(staticMenuData.insertPhoneMask);
     }
 
     // Create "Manage all aliases" link
@@ -374,6 +386,16 @@ const relayContextMenus = {
       );
       return relayAddresses;
     },
+    getPhoneMask: async () => {
+      // Note: This function only returns the mask saved in local storage.
+      //       That means that it will be out of date if this recently changed.
+      //       However, blocking the expansion of the context menu while waiting
+      //       for a GET request is not a great experience.
+      const { relayNumbers } = await browser.storage.local.get(
+        "relayNumbers"
+      );
+      return relayNumbers?.[0] ?? null;
+    },
     checkIfAnyMasksWereGeneratedOrUsedOnCurrentWebsite: async (website) => {
       const relayAddresses = await relayContextMenus.utils.getAliases();
 
@@ -519,6 +541,7 @@ browser.contextMenus.onClicked.addListener(async (info, tab) => {
   );
   const urlPremium = `${relaySiteOrigin}/premium?utm_source=fx-relay-addon&utm_medium=context-menu&utm_content=get-premium-link`;
   const urlManageAliases = `${relaySiteOrigin}/accounts/profile/`;
+  const phoneMask = await relayContextMenus.utils.getPhoneMask();
   switch (info.menuItemId) {
     case "fx-private-relay-generate-alias":
       sendMetricsEvent({
@@ -546,6 +569,26 @@ browser.contextMenus.onClicked.addListener(async (info, tab) => {
         label: "context-menu-relay-manage-aliases",
       });
       await browser.tabs.create({ url: urlManageAliases });
+      break;
+    case staticMenuData.insertPhoneMask.id:
+      sendMetricsEvent({
+        category: "Extension: Context Menu",
+        action: "click",
+        label: "context-menu-insert-phone-mask",
+      });
+      if (phoneMask) {
+        browser.tabs.sendMessage(
+          tab.id,
+          {
+            type: "fillTargetWithRelayNumber",
+            targetElementId: info.targetElementId,
+            relayNumber: phoneMask,
+          },
+          {
+            frameId: info.frameId,
+          }
+        );
+      }
       break;
   }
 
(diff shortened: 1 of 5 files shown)


src/js/other-websites/fill_relay_address.js
```diff
@@ -20,6 +20,26 @@ function fillInputWithAlias(emailInput, relayAlias) {
 
 }
 
+function fillInputWithRelayNumber(emailInput, relayNumber) {
+
+  // BUG: Duplicate fillInputWithAlias calls without proper input content
+  // The relayNumber/emailInput arguments check below is a work-around to let the duplicate call(s) fail silently.
+  // To debug, check all instances where fillInputWithRelayNumber() is being called and isolate it.
```
Vinnl, line 27: Copied from: https://github.com/mozilla/fx-private-relay-add-on/blob/21ebd978489d5cd0df8a3eb76eab0f54a8ef179e/src/js/other-websites/fill_relay_address.js#L4-L6

maxxcrawford (changes requested):
There weren't concrete testing steps, so I ran through this process on my own, and had what I think to be an acceptance criteria issue. 

- Run add-on
- Log in with non-premium account
- Go to airbnb.com and open up login modal
- Right-click phone field, see Relay context menu
- **Expected:** There should be no phone menu item bc I'm a free account
- **Actual:** The menu item, "Use phone number mask" is visible. 
  - _Note that when I click on the item, it does send a telemetry ping, but nothing is inserted, nor is any error thrown._ 
 
Doing a follow-up review with a premium-account now.

src/js/background/context-menu.js
```diff
@@ -30,6 +30,13 @@ const staticMenuData = {
     visible: true,
     contexts: ["all"],
   },
+  insertPhoneMask: {
+    id: "fx-private-relay-insert-phone-mask",
+    title: browser.i18n.getMessage("pageInputIconInsertPhoneMask"),
+    enabled: true,
```
maxxcrawford, line 36: thought (non-blocking): Not sure if we want to show this greyed out / disabled when the user does not have a phone mask set. If so, you'll want to follow the pattern like in `generateAliasDisabled` where you make another context menu entry like `insertPhoneMask`, but set `enabled: false`.

src/js/background/context-menu.js
```diff
@@ -374,6 +386,16 @@ const relayContextMenus = {
       );
       return relayAddresses;
     },
+    getPhoneMask: async () => {
+      // Note: This function only returns the mask saved in local storage.
+      //       That means that it will be out of date if this recently changed.
+      //       However, blocking the expansion of the context menu while waiting
+      //       for a GET request is not a great experience.
```
maxxcrawford, line 393: praise: Agreed wholeheartedly. Since this mask number is _semi-permanent_ for the near future, this is very appropriate.

src/js/background/context-menu.js
```diff
@@ -165,6 +172,11 @@ const relayContextMenus = {
       }
     }
 
+    const phoneMask = relayContextMenus.utils.getPhoneMask();
```
maxxcrawford, line 175: I found the issue! This is an async function.

```
await relayContextMenus.utils.getPhoneMask()
```

~Note: per my previous review, this is returning true with a free Relay account. Noting that here.~

src/js/other-websites/fill_relay_address.js
```diff
@@ -20,6 +20,26 @@ function fillInputWithAlias(emailInput, relayAlias) {
...
+  if (!emailInput || !relayNumber) {
+    return false;
+  }
+
+  // Set the value of the target field to the mask number
+  const normalisedNumber = relayNumber.number.startsWith("+1")
```
maxxcrawford, line 33: question: I know this will become an issue if we launch outside of the US. Is there a way to dynamically catch the `+1` without hardcoding?

src/js/relay.firefox.com/get_profile_data.js
```diff
@@ -107,6 +108,13 @@
 
     const siteStorageEnabled = serverProfileData[0].server_storage;
 
+    const relayNumbers = await apiRequest(relayApiUrlRelayNumbers);
+    if (Array.isArray(relayNumbers) && relayNumbers.length > 0) {
+      browser.storage.local.set({
+        relayNumbers: relayNumbers,
+      });
```
maxxcrawford, line 115: :praise: Local all the things for the performance 💪

maxxcrawford (changes requested):
Tested with a premium account without issue. Feature works great. 

I was able to debug the issue reported earlier. Great work!

src/js/background/context-menu.js
maxxcrawford, line 175:
```suggestion
    const phoneMask = await relayContextMenus.utils.getPhoneMask();
```

Vinnl: Wow! Not just identifying an issue I missed, but also debugging it to find the root cause and provide a fix - thank you so much! Fixed in b19c445

(Now if only [we had some tools](https://mozilla-hub.atlassian.net/browse/MPP-2512) that would've allowed me to catch this without relying on a very thorough human review :sweat_smile: )

src/js/background/context-menu.js
Vinnl: Thanks for the pointers. I think for now we're good, since the barrier to being able to insert a phone mask is quite a bit higher than the one to being able to generate a mask, in addition to there not being a clear path for users on how to enable it, and it not being listed in the acceptance criteria.

src/js/other-websites/fill_relay_address.js
Vinnl: Yeah we're going to have a lot of rework to do (not just here, but in the website too) if we expand beyond `+1`, and I unfortunately don't have a good answer to how to dynamically catch it. I think the best we can do when we expand is have a list of all possibile country codes...

That said, in this case it might not be too much of a problem: if the phone number doesn't start with a `+1`, it won't be stripped off, and we'll just insert the phone number including a country code. We're already accepting that this prefill will be imperfect for many possible validations. And since some forms might actually _require_ the country code, it will probably just work in different situations.

Vinnl:
Thanks for the thorough review @maxxcrawford! Added the fix in b19c445, and created https://github.com/mozilla/fx-private-relay-add-on/pull/434 to help me remember adding testing instructions in the future - sorry about that.

src/js/background/context-menu.js
maxxcrawford: _If only! 🥲_

src/js/other-websites/fill_relay_address.js
maxxcrawford: You called out the higher level answer: expanding outside of `+1` is going to be a lot of work the whole app, not just here, subtly in the add-on.

src/js/background/context-menu.js
maxxcrawford: Yeah - I know this was stripped-down MVP feature set. I figure there will be some sort of upsell surface(s) in the future. Good to go for now.

PR #436, opened 2022-11-17
Demo: Add logic for users to right-click and initiate phone call with Relay

maxxcrawford:


src/js/background/context-menu.js
```diff
@@ -37,6 +37,13 @@ const staticMenuData = {
     visible: true,
     contexts: ["all"],
   },
+  makePhoneCall: {
+    id: "fx-private-relay-call-via-relay",
+    title: "Start Relay masked call",
```
maxxcrawford, line 42: Issue (blocking): String needs to be localized

PR #452, opened 2023-01-18
Add dimension5 and dimension7 to context menu events

maxxcrawford:


Vinnl (approved):
Looks good, but I'm wondering if maybe we should run the listeners _before_ sending the event, since this adds an `await` in there? (We might also be able to get away with not even `await`ing the listeners, since we don't do anything with their return value anyway.)

src/js/background/context-menu.js
```diff
@@ -567,6 +571,8 @@ browser.contextMenus.onClicked.addListener(async (info, tab) => {
         category: "Extension: Context Menu",
         action: "click",
         label: "context-menu-relay-manage-aliases",
+        dimension5: await getBrowser(),
+        imension7: "add-on",
```
Vinnl, line 575: Happened to see this just now, thought I'd call it out just in case:

```suggestion
        dimension7: "add-on",
```

maxxcrawford:
@Vinnl Great catch. I moved the await to the top of the function, rather than calling on every switch. Trying out it with await'less.

Issue #454, opened 2023-01-21
Multiple Relay Icon Bug


ACK-J:
Browse to https://www.tari.com/ and view the icon bug as shown below

![image](https://user-images.githubusercontent.com/60232273/213872024-fb13ca5d-b613-428a-a927-248ed9923f4d.png)

Issue #447, opened 2022-12-20
Textbox that Relay Injects Looks Broken


dshin-moz:
STR:

1. Go to [Sony PSN page](https://www.playstation.com/en-ca/playstation-network/)
2. Click "Sign In"  
3. Examine the "Sign-In ID" field

Expected: Relay-injected sign in field is still styled as per the website's CSS
![without_relay](https://user-images.githubusercontent.com/102040459/208755895-870063bd-1456-4c4f-8471-e94f75c864d6.png)

Actual: Looks like a default-styled input field. Launching on fresh profile, I see the input field is styled as expected.

![with_relay](https://user-images.githubusercontent.com/102040459/208755900-d803ba1a-39d6-4780-9f06-b667166de9f6.png)

lotas:
I noticed similar problem on the https://canarytokens.org/generate website:

![image](https://user-images.githubusercontent.com/83861/214114116-3bd26396-60e6-425d-836b-57e6d9251706.png)

groovecoder:
@lotas that image is a canary token itself, isn't it? 😛

Issue #400, opened 2022-09-22
breaks login to upwork.com


designakt:
Disables the email field so that no input is possible
<img width="540" alt="Screenshot 2022-09-22 at 08 54 19" src="https://user-images.githubusercontent.com/8145891/191678687-60f9498d-d533-45c4-8460-6df83a7670f9.png">

login url: https://www.upwork.com/ab/account-security/login

maxxcrawford:
@designakt 

If you want to turn this feature off, you can through the settings panel in the add-on:

1. Click on the Relay icon in the URL bar
    <img width="139" alt="image" src="https://user-images.githubusercontent.com/2692333/194422138-0e28a6b0-6be2-4ccb-a54f-4a30ee2c4226.png">
2. Click on the Gear ⚙️ icon
    <img width="385" alt="image" src="https://user-images.githubusercontent.com/2692333/194422200-795e242c-e206-437b-a530-34cdd2237391.png">
3. Click the first toggle item
    <img width="383" alt="image" src="https://user-images.githubusercontent.com/2692333/194422219-69ad7c32-e582-4f95-81e5-f5da93f96258.png">

If you do encounter a field that you want to generate an email mask on, you can right/alt-click and use the Firefox Relay link in the context menu. Hope that helps for now!

designakt:
Thanks ! That is certainly better then turning the extension off completely.

Issue #397, opened 2022-09-09
Overlay changes layout in Pontoon profile page


flodolo:
If you have a Pontoon profile, just visit the [settings page](https://pontoon.mozilla.org/settings/) and you'll notice the email field out of line.

<img width="651" alt="Schermata 2022-09-09 alle 15 00 08" src="https://user-images.githubusercontent.com/3644868/189355982-d0da930f-771b-42fa-afb2-f72d3b841fc2.png">

PR #462, opened 2023-01-27
Change icon: overlaid button to background image

### LINKED ISSUE #22: High z-index value for email field wrapper causes stacking issues

Vinnl:
This PR fixes #22, #80, #227, #253, #360, #358, #394, #397, #400, #429, #447, #454, #497, #509.

(For anyone coming across this PR: we'll still need to do some extensive testing to make sure this doesn't break commonly-used websites, so please be patient — it may still take a while to land.)

# New feature description

This applies a new method of injecting the Relay icon that I prototyped a while ago; a helpful description of this method can be found [here](https://github.com/mozilla/fx-private-relay-add-on/issues/22#issuecomment-1372445245), by @czlucius who independently landed on the same approach.

With this method, we don't have to change the stacking order of the website, and thus should avoid a lot of issues where we break the website we're injecting in. It did require some extra work to stay accessible; we still inject a button that is visually hidden, but accessible to screen reader users and people who navigate using the keyboard. I've also modified the hover/focus state styling to make it easier. Scrolling with the menu open is no longer possible; I can probably re-enable that with a re-positioning scroll listener, but I feel like the user experience is probably fine if you dismiss the menu first, and should be better for performance.

Note that I carried over the code in src/js/other-websites/icon.js from the experimental extension I prototyped this in, so the code style might be a little different from the rest of the codebase - the experimental extension had TypeScript and helpers I've to rip out.

# Screenshot (if applicable)

image

[Screencast from 2023-01-27 11-26-42.webm](https://user-images.githubusercontent.com/4251/215064658-b9d7c905-7bca-4d28-af69-275e0a45120e.webm)

image

image

image

image

image

image

image

image

# How to test

Visit websites with the Relay icon, and make sure that everything looks and works OK.

# Checklist

- [x] All acceptance criteria are met.
- [ ] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).


src/js/other-websites/icon.js
```diff
@@ -0,0 +1,427 @@
...
+      typeof message.newRelayAddressResponse === "object" &&
+      (
+        typeof message.newRelayAddressResponse?.full_address === "string" ||
+        typeof message.newRelayAddressResponse?.address === "string"
+      )
+    ) {
+      // When generating a new mask, `newRelayAddressResponse` contains a mask object
+      // with the `full_address` property. When selecting an existing mask,
+      // however, it only has the properties `address` and `currentDomain`, but
+      // `address` does contain the full address (i.e. including @mozmail.com):
```
Vinnl, line 191: I miss TypeScript :(

Vinnl:
Added two more features based on PM feedback:

- Scrolling is no longer disabled, and the menu moves along with the icon on scroll: [eb2c59c](https://github.com/mozilla/fx-private-relay-add-on/pull/462/commits/eb2c59c3ee6e23b29e15e7e226ea382320442290)
- If there's not enough space for the menu below the icon, and there is enough space on top, move the menu to the top: [e5cfe23](https://github.com/mozilla/fx-private-relay-add-on/pull/462/commits/e5cfe23c4f37d18a19288fa87e88f6efc1b72669)

src/js/other-websites/icon.js
```diff
@@ -0,0 +1,452 @@
...
+
+/**
+ * @param {MouseEvent} event 
+ * @returns void
+ */
+function onEmailInputClick(event) {
+  const target = event.target;
+  if (!(target instanceof HTMLInputElement)) {
+    return;
+  }
```
maxxcrawford, line 147: Praise: Typescript isn't here, but this type check still is. 💓

src/js/other-websites/icon.js
```diff
@@ -0,0 +1,452 @@
...
+  underlay.style.position = "fixed";
+  underlay.style.inset = "0";
+  underlay.style.zIndex = "99999999";
+  underlay.style.backgroundColor = "rgba(0, 0, 0, 0.1)";
+  const existingIframe = underlay.querySelector("iframe");
+  const iframe = existingIframe ?? document.createElement("iframe");
+  iframe.src = browser.runtime.getURL("inpage-panel.html");
+  iframe.style.position = "absolute";
+  iframe.style.width = `${RELAY_INPAGE_MENU_WIDTH}px`;
+  iframe.style.maxWidth = `95vw`;
```
maxxcrawford, line 178: Per my comment at the top of this review, I expect how the iframe is being initialized is where we would set the tip/rounded corners. 

image

src/js/other-websites/icon.js
```diff
@@ -0,0 +1,452 @@
...
+  };
+  browser.runtime.onMessage.addListener(onUpdateIframeHeight);
+
+  /**
+   * @param {MouseEvent} _event 
+   */
+  const onUnderlayClick = (_event) => {
+    closeMenu();
+  };
+  underlay.addEventListener("click", onUnderlayClick);
```
maxxcrawford, line 240: Question/thought: Do we need any [`touchstart`](https://developer.mozilla.org/en-US/docs/Web/API/Element/touchstart_event) support for the Android version of this add-on?

src/js/other-websites/icon.js
```diff
@@ -0,0 +1,452 @@
...
+ * @param {HTMLInputElement} element
+ */
+function setHoverStyles(element) {
+  // Styles to add hover effects (background glow and a pointer cursor),
+  // while preserving the existing background styles for this element:
+  const elementStyles = getComputedStyle(element);
+  const hoverBackgroundStyle =
+    elementStyles.backgroundImage +
+    `, linear-gradient(to left, #f68fff ${
+      getPaddingRight(element) * 2 + 25
```
maxxcrawford, line 369: Suggestion: Can you move `#f68fff` to a variable at the top of the file? I feel like this is hidden/buried CSS in this file that we'd want to better surface.

maxxcrawford: Extra suggestion: Move all/any CSS attributes to the top of the file.

maxxcrawford (approved):
This PR has been a long time coming! Thanks, @Vinnl (and @czlucius too!) We'll need to do some extensive testing across sites to confirm there's no major breakage, but this so less invasive from our current approach. 

Approving with some nits/suggestions/questions! 

One minor UI inconsistency to call-out: 
The in-page menu has both rounded corners and a small tooltip pointing towards the input: 
image

image

I called out in the code where I believe it would be set, if that helps?

src/js/other-websites/email_detector.js
```diff
@@ -89,23 +89,25 @@ const email_detector_ruleset = ruleset([
...
-        }
-    }
+  // Then run ruleset and return detected fields
+  const detectedInputs = email_detector_ruleset
+    .against(domRoot)
+    .get("email")
+    .filter((node) => node.scoreFor("email") > 0.5)
+    .map((node) => node.element);
+
+  return typeEmailInputs.concat(detectedInputs);
```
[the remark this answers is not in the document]
Vinnl: It was a generator function, where that didn't add much value as far as I could see (checking the commit log), while making it hard to pass around the list of email inputs. And then the rest was to make sure I actually understood what was going on here, and writing it myself was a good way to do so :) But happy to change it back to for loops if you prefer, now that I know what's going on :)

src/js/other-websites/fill_relay_address.js
```diff
@@ -54,6 +54,7 @@ if (!browser.menus) {
 }
 
 browser.runtime.onMessage.addListener((message, _sender, _response) => {
+  // These messages are sent by the context menu, not the in-page popup:
   if (message.type === "fillTargetWithRelayAddress") {
```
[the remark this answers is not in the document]
Vinnl: Hahaha, feels like we've gone through the same pain :sweat_smile:

src/js/other-websites/icon.js
Vinnl: Hmm crap, in my spike I added this icon inside the iframe. I'll have to do some rework to be able to add it here, since you can't add `::before` elements (or other children) to an iframe, so all the positioning will have to be redone to apply to a wrapper div. Hopefully I can find some time for that next week.

(Btw, this PR does already have rounded corners, and as far as I can see they're the same size -4px- as the current implementation?)

src/js/other-websites/icon.js
Vinnl: Moved all colors in 3722328. Not sure if there were additional properties you were thinking of? (Note that most styles are dynamically generated based on existing styles for a particular input element, so I can't just move them to the top as hardcoded strings.)

src/js/other-websites/icon.js
Vinnl: I don't _think_ so, as [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event#safari_mobile) mentions a bug in old Safari versions for `click` events, implying that they're supposed to fire, and [StackOverflow says so](https://stackoverflow.com/a/37273344). But I'm not sure how to verify - do you know if there's an easy way to test `.xpi` files on Firefox mobile?

src/js/other-websites/icon.js
Vinnl: OK, I've struggled with making this appear properly for a bit now (in particular, getting it to take up the right space and have a consistent size), and I can't seem to make it work. My proposal would be to go without the chevron for now - it's relatively clear that the icon is related to the icon, so I think not having it for now would be better than having the current buggy icon implementation. I'm going to ask QA to test this.

Vinnl:
@maxxcrawford I added cbde420 that fixes an issue where the spinner would never stop in Chrome - the commit message should explain more. Could you review that too, please? With that, I think that this might be ready to go.

Vinnl:
Rebased to clean up commits and resolve confilcts - specifically, I reapplied https://github.com/mozilla/fx-private-relay-add-on/pull/499 in e06c3ae.

There were also https://github.com/mozilla/fx-private-relay-add-on/commit/056a91e6259e033cb005d48898b55f7b35d89946#diff-6d657968b51497d6080c30b3704a5e0cbac32a4707fb358dd33955f628c5d192 and https://github.com/mozilla/fx-private-relay-add-on/commit/9b31315bb3db9bc93e558b33aad5f7b96b3d9072#diff-59a5db4e9a139708113dc0d0e60807e300b556f19ce611a5ffaa4a764c83f14b, but as far as I could see those were concerned with setting a particular height for the in-page popup, which IIRC should get determined automatically after this PR.

hecerinc:
@Vinnl just checking if there's plans to merge this any time soon. Looks like you have this approved? Is there a timeline set for this?

Vinnl:
@hecerinc I moved to work on Monitor and handed this over to other folks, but it looks like they haven't gotten around to releasing it yet. I'll bug some folks again to see if we can get this released :)

PR #479, opened 2023-03-10
Process GitHub Actions deprecations

Vinnl:
I just noticed a bunch of deprecations in the prerelease job:

image

So I bumped the Node version to 16, and incorporated this change for output variables: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


maxxcrawford (approved):
LGTM

PR #486, opened 2023-03-29
[MPP-1708] Addon Toolbar Refactor: Additional QA Fixes

maxxcrawford:
## Summary

Issues:
- https://mozilla-hub.atlassian.net/browse/MPP-3006
- https://mozilla-hub.atlassian.net/browse/MPP-2863

## Testing 

**MPP-2863:**
- Run the add-on, Log in
- Open panel, Navigate to Settings > Report Issue
- **Expected:** Form is visible but the submit button is disabled
  - If on an external (non `about:` or `chrome://`) webpage, the first field should be prefilled with current URL hostname
- Check at least checkbox that isn't OTHER
- **Expected:** Submit button is enabled and the form can be submitted
- Check the OTHER Checkbox
- **Expected:** Submit button is now disabled and the form cannot be submitted
- Fill out OTHER Input field
- **Expected:** Submit button is enabled and the form can be submitted

**MPP-3006**
- Run the add-on, Log-in with premium account
- Make multiple masks so that you can scroll through the masks 
- While scrolled down the list, hover over any navigation item (News, etc) 
- **Expected:** The navigation item tool tip should be above the main content

## Screenshots

MPP-3006: 
image

### DIFF
--- a/src/js/background/background.js
+++ b/src/js/background/background.js
@@ -113,12 +113,15 @@ async function postReportWebcompatIssue(description) {
     user_agent: description.user_agent
   };
 
-  await fetch(reportWebCompatResponse, {
+  const newReportedIssueFetch = await fetch(reportWebCompatResponse, {
     mode: "same-origin",
     method: "POST",
     headers: headers,
     body: JSON.stringify(apiBody),
   });
+
+  return newReportedIssueFetch;
+
 }
 
 async function storeRuntimeData(opts={forceUpdate: false}) {  
@@ -509,7 +512,7 @@ browser.runtime.onMessage.addListener(async (m, sender, _sendResponse) => {
       response = await makeRelayAddress(m.description);
       break;
     case "postReportWebcompatIssue":
-      await postReportWebcompatIssue(m.description);
+      response = await postReportWebcompatIssue(m.description);
       break;
     case "openRelayHomepage":
       browser.tabs.create({
--- a/src/js/background/background.js
+++ b/src/js/background/background.js
@@ -96,9 +96,11 @@ async function patchMaskInfo(method = "PATCH", id, data, opts=null) {
 
 async function postReportWebcompatIssue(description) {
   const { relayApiSource } = await browser.storage.local.get("relayApiSource");  
+  
   if (!relayApiSource) {
     return;
   }
+
   const headers = await createNewHeadersObject({auth: true});
   const reportWebCompatResponse = `${relayApiSource}/report_webcompat_issue`;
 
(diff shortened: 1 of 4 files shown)


src/js/background/background.js
```diff
@@ -111,16 +113,12 @@ async function postReportWebcompatIssue(description) {
     user_agent: description.user_agent
   };
 
-  const newReportedIssueResponse = await fetch(reportWebCompatResponse, {
+  await fetch(reportWebCompatResponse, {
     mode: "same-origin",
     method: "POST",
     headers: headers,
     body: JSON.stringify(apiBody),
   });
```
lloan, line 121: Question (non-blocking) ln 116 in background.js - you have it waiting for fetch to complete (post request) but never return a response regarding if it was successful or not. And on ln 512, you have it just waiting for the outcome of the function, `postReportWebcompatIssue` but it doesn't return anything after it completes the call. We might want to be a little bit more intentional with our code, no? So we know what to expect when we do other calls in the future (response of some sort - not just a side effect that returns silence).

maxxcrawford: Ugh. I think something is wrong with the API from the web-app side. Gonna check with back-end tomorrow.

image

maxxcrawford: Per jwhitlock, this is intended. Submitting to this endpoint only returns an HTTP status code. There's no JSON response, etc.

src/js/popup/popup.js
```diff
@@ -26,6 +26,19 @@
...
+          popup.panel.webcompat.resetForm();
+
+          // Remove event listeners
+          const panel = document.getElementById('webcompat-panel');
+          const clonedPanel = panel.cloneNode(true);
+          panel.parentNode.replaceChild(clonedPanel, panel);
+
+          // Set back button listeners again for webcompat
+          const webcompatBackButton = document.querySelector("#webcompat-panel .fx-relay-panel-header-btn-back");
+          webcompatBackButton.addEventListener("click", popup.events.backClick, false);
```
maxxcrawford, line 40: Because we return the user back to the settings page after submitting the form, we have to do a heavy reset of the form so they can fill it out again. 

The `replaceChild(clonedPanel, panel);` part is a [hacky way](https://stackoverflow.com/questions/65495153/remove-all-event-listeners-from-an-element-javascript) to remove all anonymous event listeners on an element.

lloan: Not sure if this helps since you already have a working way, but couldn't we also do something like: 

```
// first we get ref. to element in question 
const element = document.getElementById('webcompat-panel');

// call getEventListeners to get an object representing all the event listeners on the element
const listeners = getEventListeners(element);

for (const eventType in listeners) {
  const eventListeners = listeners[eventType];

// iterate through each event type (e.g. 'click', 'mouseenter', etc.) in the listeners object, 
// and for each event type we iterate through the array of event listeners for that type. 
  for (const eventListener of eventListeners) {

    // If an event listener doesn't have a name property (which means it's an anonymous function)
    if (!eventListener.name) {

      // we remove it 
      element.removeEventListener(eventType, eventListener.listener);
    }
  }
}```

lloan (approved):
Verified MPP-2863 works as intended. 

Verified MPP-3006 works as intended. 

Tested on FF. 

I left a comment - non-blocking since you already have a working solution, but I did add a different approach.

lloan (approved):
Thanks for the comments for the magic numbers!

Should we open an issue for the FIXME so we don't forget about it, linking to the fix me code line?

PR #488, opened 2023-04-06
Only suggest inserting masks in text fields

Vinnl:
The context menu entries to generate/reuse a mask only really make sense if the user is right-clicking an editable field. (Ideally even just email address fields, but we don't have an API that allows us to detect that.) @maxxcrawford let me know if there was a reason you initially had these set to `all`. See https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus/ContextType.

This PR fixes MPP-3014.

How to test: right-click an email input - nothing should have changed. Right-click anything else, e.g. the Relay icon in the toolbar, and you should not have options to generate or reuse a mask.

- [x] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).

### DIFF
--- a/src/js/background/context-menu.js
+++ b/src/js/background/context-menu.js
@@ -8,21 +8,21 @@ const staticMenuData = {
   existingAlias: {
     type: "radio",
     visible: true,
-    contexts: ["all"],
+    contexts: ["editable"],
   },
   generateAliasEnabled: {
     id: "fx-private-relay-generate-alias",
     title: browser.i18n.getMessage("pageInputIconGenerateNewAlias_mask"),
     enabled: true,
     visible: true,
-    contexts: ["all"],
+    contexts: ["editable"],
   },
   generateAliasDisabled: {
     id: "fx-private-relay-generate-alias",
     title: browser.i18n.getMessage("pageInputIconGenerateNewAlias_mask"),
     enabled: false,
     visible: true,
-    contexts: ["all"],
+    contexts: ["editable"],
   },
   manageAliases: {
     id: "fx-private-relay-manage-aliases",
@@ -35,7 +35,7 @@ const staticMenuData = {
     title: browser.i18n.getMessage("pageInputIconInsertPhoneMask"),
     enabled: true,
     visible: true,
-    contexts: ["all"],
+    contexts: ["editable"],
   },
   upgradeToPremium: {
     id: "fx-private-relay-get-unlimited-aliases",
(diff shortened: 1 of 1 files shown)


src/js/background/context-menu.js
```diff
@@ -35,7 +35,7 @@ const staticMenuData = {
     title: browser.i18n.getMessage("pageInputIconInsertPhoneMask"),
     enabled: true,
     visible: true,
-    contexts: ["all"],
+    contexts: ["editable"],
```
maxxcrawford, line 38: suggestion (non-blocking): I would recommend leaving this one as is. Having a way to open Relay anywhere can be useful. 

```suggestion
    contexts: ["all"],
```

src/js/background/context-menu.js
```diff
@@ -55,13 +55,13 @@ const staticMenuData = {
       "pageInputIconUseExistingAliasFromTheSite_mask"
     ),
     visible: true,
-    contexts: ["all"],
+    contexts: ["editable"],
```
maxxcrawford, line 58: suggestion (non-blocking): I would recommend leaving this one as is. Having a way for users to upgrade may be useful? This may also seem like too upgrade-focused. LMK your thoughts.  

```suggestion
    contexts: ["all"],
```

maxxcrawford (approved):
I added two questions, both non-blocking. I do think we should keep the link for managing masks everywhere (especially in the context of clicking on the icon in the toolbar) 

image

src/js/background/context-menu.js
Vinnl: GitHub's diff collapsing makes this looks like this affects the `manageAliases` entry, but it's actually `insertPhoneMask`:

[Screencast from 2023-04-13 10-53-26.webm](https://user-images.githubusercontent.com/4251/231707777-67ec44e4-e4fa-498a-bc19-146006a59809.webm)

"Manage aliases" will indeed still be available everywhere.

(I was so confused when I went to make this change :sweat_smile:

src/js/background/context-menu.js
Vinnl: And here too, this is actually `useExistingAliasFromWebsite` rather than `upgradeToPremium` :)

Vinnl:
@maxxcrawford requested another review - I'm assuming this is good to merge, but because GH's diff display is confusing, might be good to take another look to be sure.

Issue #495, opened 2023-04-16
bad perceived performance of opening the panel


cadeyrn:
STR:

1. Click the Firefox Relay button in the extension menu

Actual:

1. First you see an empty panel with the header row and the big blue button  to create a custom mask on the bottom
2. Then the blue button moves a bit up and a link below the button appears to create a random mask
3. Then the existing masks and the filter field appears

It takes ~ 2 seconds until everything is there, every time you press the button. This gives the impression of a really bad performance.

Some ideas to consider:

1. Is there something that could be cached so that it's fast from the second time you open the panel at least?
2. Is it possible to let the blue button and the link for creating a random mask appear at the same time so that nothing moves? Even if it still takes the same time until everything is there it would improve the perceived performance if less visible steps are involved and no content moves.

---

add-on version: 2.7.0

PR #498, opened 2023-04-19
[MPP-3063] Reduce panel load time

maxxcrawford:
## Summary 
This PR reduces the load time when a user is logged in and opens the panels to view their masks. 

This fixes: 
- [MPP-3063](https://mozilla-hub.atlassian.net/browse/MPP-3063)
- #495 

TODO
- [x] Remove extra API call to get latest masks
  - [x] Confirm other pages do not make the same call. Use local storage to power other panels (Stats)
- [x] ~Build initial masks lists on local (stale) date~ Disregarding for now
  - [x] ~Update masks accordingly when API call finishes loading (add new masks, update status of other masks)~ 
- [x] Move "save to local storage" logic when API is called for masks to background scripts
- [x] Edge case: Update mask view accordingly when API call finishes loading if on empty (no mask) state

## Testing

For context, compare these actions to the production release with the same account. 

- If possible, use personal Relay account with 100+ accounts (May have to run `$ npm run config:prod`) 
- Log into account
- Open panel
- **Expected:** Panel should load within ~1s and the loading state should be visible until the data is shown
- Open stats
- **Expected:** Panel should load INSTANTLY with data 

## Screenshots

TBD

### DIFF
--- a/src/js/popup/popup.js
+++ b/src/js/popup/popup.js
@@ -476,85 +476,14 @@
             const masks = await popup.utilities.getMasks(getMasksOptions);
             
             const maskList = document.querySelector(".fx-relay-mask-list");
+
             // Reset mask list
             maskList.textContent = "";
-
-            masks.forEach(mask => {
-              const maskListItem = document.createElement("li");
-
-              // Attributes used to power search filtering
-              maskListItem.setAttribute("data-mask-address", mask.full_address);              
-              maskListItem.setAttribute("data-mask-description", mask.description ?? "");
-              maskListItem.setAttribute("data-mask-used-on", mask.used_on ?? "");
-              maskListItem.setAttribute("data-mask-generated", mask.generated_for ?? "");
-              
-              maskListItem.classList.add("fx-relay-mask-item");
-
-              const maskListItemNewMaskCreatedLabel = document.createElement("span");
-              maskListItemNewMaskCreatedLabel.textContent = browser.i18n.getMessage("labelMaskCreated");
-              maskListItemNewMaskCreatedLabel.classList.add("fx-relay-mask-item-new-mask-created");
-              maskListItem.appendChild(maskListItemNewMaskCreatedLabel);
-              
-              const maskListItemAddressBar = document.createElement("div");
-              maskListItemAddressBar.classList.add("fx-relay-mask-item-address-bar");
-
-              const maskListItemAddressWrapper = document.createElement("div");
-              maskListItemAddressWrapper.classList.add("fx-relay-mask-item-address-wrapper");
-
-              const maskListItemLabel = document.createElement("span");
-              maskListItemLabel.classList.add("fx-relay-mask-item-label");
-              maskListItemLabel.textContent = mask.description;
-              
-              // Append Label if it exists
-              if (mask.description !== "") {
-                maskListItemAddressWrapper.appendChild(maskListItemLabel);
-              }
-              
-              const maskListItemAddress = document.createElement("div");
-              maskListItemAddress.classList.add("fx-relay-mask-item-address");
-              maskListItemAddress.textContent = mask.full_address;
-              maskListItemAddressWrapper.appendChild(maskListItemAddress);
-
-              // Add Mask Address Bar Contents 
-              maskListItemAddressBar.appendChild(maskListItemAddressWrapper);
-
-              const maskListItemAddressActions = document.createElement("div");
-              maskListItemAddressActions.classList.add("fx-relay-mask-item-address-actions");
-
-              const maskListItemCopyButton = document.createElement("button");
-              maskListItemCopyButton.classList.add("fx-relay-mask-item-address-copy");
-              maskListItemCopyButton.setAttribute("data-mask-address", mask.full_address);
-
-              const maskListItemCopyButtonSuccessMessage = document.createElement("span");
-              maskListItemCopyButtonSuccessMessage.textContent = browser.i18n.getMessage("popupCopyMaskButtonCopied");
-              maskListItemCopyButtonSuccessMessage.classList.add("fx-relay-mask-item-address-copy-success");
-              maskListItemAddressActions.appendChild(maskListItemCopyButtonSuccessMessage);
-              
-              maskListItemCopyButton.addEventListener("click", (e)=> {
-                e.preventDefault();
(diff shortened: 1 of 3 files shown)


src/js/popup/popup.js
```diff
@@ -343,10 +343,12 @@
             }, false);
           
           // Get masks and determine what to display
-          const masks = await popup.utilities.getMasks(getMasksOptions);
+          const { relayAddresses } = await browser.storage.local.get(
+            "relayAddresses"
+          );
```
maxxcrawford, line 348: Context: This was one logic point where where we switched to local storage instead of waiting for the API results.

src/js/popup/popup.js
```diff
@@ -577,6 +507,86 @@
               popup.panel.masks.search.init();
             }
 
+            // Remove loading state
+            document.body.classList.remove("is-loading");
```
maxxcrawford, line 511: Context: Moved this to the end of the `buildMasksList()` function. Previously, we were removing the loading state too early.

src/js/popup/popup.js
```diff
@@ -343,10 +343,12 @@
...
           
           // Get masks and determine what to display
-          const masks = await popup.utilities.getMasks(getMasksOptions);
+          const { relayAddresses } = await browser.storage.local.get(
+            "relayAddresses"
+          );
 
           // If no masks are created, show onboarding prompt
-          if (masks.length === 0) {
+          if (relayAddresses.length === 0) {
```
maxxcrawford, line 351: TODO: There may be an edge case where a user initially has no masked created when using the add-on, and then creates a mask elsewhere (other browser, etc) and they get the empty state when they should see their current masks. 

This may be the one spot to live/refresh based on new data.

src/js/popup/popup.js
```diff
@@ -475,85 +476,14 @@
...
-              // TODO: Add toggle button back
-              // maskListItemAddressActions.appendChild(maskListItemToggleButton);
-
-              maskListItemAddressBar.appendChild(maskListItemAddressActions);
-              maskListItem.appendChild(maskListItemAddressBar);
-              maskList.appendChild(maskListItem);
+            
+            // Generate and append each mask item to the mask list
+            masks.forEach( mask => {
+                const maskListItem = popup.panel.masks.utilities.buildMaskListItem(mask);
```
maxxcrawford, line 485: Context: This function was getting really big. Broke out the `buildMaskListItem` iteration to make it easier to read.

src/js/popup/popup.js
```diff
@@ -577,6 +507,86 @@
...
+            maskListItemToggleButton.classList.add("fx-relay-mask-item-address-toggle");
+            maskListItemToggleButton.addEventListener("click", ()=> {
+              // TODO: Add Toggle Function
+            }, false);
+            maskListItemToggleButton.setAttribute("data-mask-id", mask.id);
+            maskListItemToggleButton.setAttribute("data-mask-type", mask.mask_type);
+            maskListItemToggleButton.setAttribute("data-mask-address", mask.full_address);
+
+            // TODO: Add toggle button back
+            // maskListItemAddressActions.append(maskListItemToggleButton);
```
maxxcrawford, line 585: Disregard this TODO. It is part of #473

src/js/popup/popup.js
```diff
@@ -1240,10 +1251,15 @@
...
-            return await browser.runtime.sendMessage({
+            const masks = await browser.runtime.sendMessage({
               method: "getAliasesFromServer",
               options,
             });
+
+            await browser.storage.local.set({"relayAddresses": masks});
+
+            return masks;
+
```
maxxcrawford, line 1262: The crux of the PR. Whenever we query `getAliasesFromServer`, we store it locally as well. 

TODO: Move this logic to background script so it happens anytime it happens from anywhere.

maxxcrawford (commented):
Self-review. Came up with a few TODOs!

src/js/popup/popup.js
maxxcrawford: So...we already do this. 

https://github.com/mozilla/fx-private-relay-add-on/blob/main/src/js/background/background.js#L60

Removed in 7fa9c62db3928021dc827f160cdaabda3bfed543

src/js/popup/popup.js
maxxcrawford: Revised in dffd2674682978a1b97e460c2b67caf308196259

Moved this entire logic block to `buildMasksList()`

I ran some tests locally if a user creates a mask else where and this loads. It refreshes without issue. I also tested the inverse (local storage has one mask, but user deletes it elsewhere). All good!

src/js/popup/popup.js
maxxcrawford: Revised this placement in dffd2674682978a1b97e460c2b67caf308196259. 

However, we have still reduced the extra call(s) to the API, so the comment still stands.

src/js/popup/popup.js
```diff
@@ -577,6 +499,86 @@
...
+              setTimeout(() => {
+                maskListItemCopyButtonSuccessMessage.classList.remove("is-shown")
+              }, 1000);
+            }, false);
+            maskListItemAddressActions.append(maskListItemCopyButton);
+
+            const maskListItemToggleButton = document.createElement("button");
+            maskListItemToggleButton.classList.add("fx-relay-mask-item-address-toggle");
+            maskListItemToggleButton.addEventListener("click", ()=> {
+              // TODO: Add Toggle Function
```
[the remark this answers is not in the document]
maxxcrawford: https://github.com/mozilla/fx-private-relay-add-on/pull/498#discussion_r1172651338

src/js/popup/popup.js
```diff
@@ -475,85 +476,14 @@
...
-
-              // TODO: Add toggle button back
-              // maskListItemAddressActions.appendChild(maskListItemToggleButton);
-
-              maskListItemAddressBar.appendChild(maskListItemAddressActions);
-              maskListItem.appendChild(maskListItemAddressBar);
-              maskList.appendChild(maskListItem);
+            
+            // Generate and append each mask item to the mask list
+            masks.forEach( mask => {
```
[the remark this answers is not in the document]
maxxcrawford: No! This is the API call we want to keep. Good callout.

PR #521, opened 2023-09-19
MPP-3431 & MPP-3432 - CSAT bug fixes 

lloan:
This PR fixes MPP-3431 MPP-3432 

# New feature description

Two changes:

First change: change the ‘x’ icon in that purple banner to be a ‘>’ arrow icon. Heres why: So I think that this would match the current behavior and intention of wanting users to get 1 layer deeper before dismissing the option to take the survey all together, which is what the ‘x’ at the top level would do. So at least this solution with replacing the ‘x’ with an '>’ arrow would indicate to the user that they can drill one layer down and see what this message/alert is all about then THEN chose to dismiss it if they wish.

Second change:  allow the user to select/toggle between the different options.

# Screenshot (if applicable)
image

image

 
# How to test
 
- Close button is now a right chevron that takes you to the survey as well.
- You can no change your selection (satisfaction level) and the survey link is updated. 

# Checklist

- [ ] l10n changes have been submitted to the l10n repository, if any.
- [ ] All acceptance criteria are met.
- [ ] All UI revisions follow the [coding standards](https://github.com/mozilla/fx-private-relay/blob/main/docs/coding-standards.md), and use Protocol tokens where applicable (see `/frontend/src/styles/tokens.scss`).
- [ ] Commits in this PR are minimal and [have descriptive commit messages](https://chris.beams.io/posts/git-commit/).


src/popup.html
```diff
@@ -86,8 +86,8 @@ <h4 class="i18n-content" data-i18n-message-id="popupSignUpPanelWelcome"></h4>
             <!-- CSAT Survey -->
             <div class="fx-relay-csat-survey-link-container is-hidden" role="alert" aria-labelledby="surveyLabel" aria-hidden="true">
                 <button id="survey-label" class="i18n-content fx-relay-csat-survey-link" data-i18n-message-id="popupSurveyLink" tabindex="0">
-                <button class="fx-relay-csat-survey-close-icon" aria-label="Close survey" tabindex="0">
-                    <img class="i18n-alt-tag" data-i18n-message-id="close" alt="Close icon" src="/icons/nebula-close.svg">
+                <button class="fx-relay-csat-survey-view-icon" aria-label="view survey" tabindex="0">
+                    <img class="i18n-alt-tag" data-i18n-message-id="view" alt="view icon" src="/icons/chevron-right.svg">
```
Vinnl, line 90: Couple of things jump out on this `aria-label` and `alt`:

1. Shouldn't they be localised?
2. If "view icon" was read out to me, I wouldn't quite know what to do :sweat_smile: 
3. I think an `aria-label` is redundant; AFAIK the accessible name of a button is its contents, which is the image, of which the accessible name should be the `alt` text.

src/css/popup.css
```diff
@@ -695,9 +695,8 @@ sign-up-panel::after {
   font-size: var(--fontSizeBodyXs);
   display: flex;
   justify-content: center;
-  gap: var(--spacingSm);
+  gap: var(--spacingLg);
   margin-bottom: var(--spacingSm);
-  align-items: center;
```
Vinnl, line 700: Judging by the screenshot, doesn't removing this cause "Help us improve Relay" to be positioned slightly downwards, as opposed to in the middle of the message bar?

Vinnl (approved):
I haven't tried this myself, but the code mostly looks sensible to me, bar one question and one a11y remark that was already present in the old code.

src/css/popup.css
lloan: image

src/popup.html
lloan: It looks like other images leave `alt=""` empty. I reused the string by the text in the banner:

image

The chevron does the exact same thing as the text. It used to be a close button, but now it just drills into the survey, going into the next panel.

PR #523, opened 2023-09-20
(WIP) Prototype of Custom Setting that offsets Relay icon on email inputs to make room for other add-ons (Password Managers, etc).

maxxcrawford:
This is not feature complete and would require lots of QA across a few "supported" add-ons.

In addition to this PR, there's another angle to come from: The "management" permission (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/management)

This permission should be optional, but would allow us to detect if any other email-input add-ons were installed and update our positions accordingly. I recommend making it optional/opt-in. It's a pretty pervasive permission.

### DIFF
--- a/src/js/other-websites/icon.js
+++ b/src/js/other-websites/icon.js
@@ -19,8 +19,15 @@ async function run() {
   if (!inputIconsAreEnabled) {
     return;
   }
+
+  const pwdMgrCompat = await isPwdMgrCompatEnabled();
+
+  if (pwdMgrCompat) {
+    // TODO: Set RELAY_ICON_WIDTH width, hover state calcs
+  }
+
   const emailInputs = findEmailInputs(document);
-  wireUpInputs(emailInputs);
+  wireUpInputs(emailInputs, pwdMgrCompat);
 
   const potentialNewInputObserver = new MutationObserver((mutations) => {
     mutations.forEach((mutation) => {
@@ -30,7 +37,7 @@ async function run() {
       for (const addedNode of mutation.addedNodes) {
         if (addedNode instanceof HTMLElement) {
           const addedEmailInputs = findEmailInputs(addedNode);
-          wireUpInputs(addedEmailInputs);
+          wireUpInputs(addedEmailInputs, pwdMgrCompat);
         }
       }
     });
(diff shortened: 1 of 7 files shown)


src/js/popup/popup.js
```diff
@@ -1775,6 +1776,36 @@
...
+          stylePrefToggle(pwdMgrCompat);
+        }
+
+        pwdMgrCompatPrefToggle.addEventListener("click", async(e) => {
+          const pwdMgrCompatPreference = e.target.dataset.pwdMgrCompatPreference;
+          await browser.storage.local.set({ "pwdMgrCompat" : pwdMgrCompatPreference });
+          stylePrefToggle(pwdMgrCompatPreference);
+        });
+        
+      },
```
maxxcrawford, line 1808: TODO: This function, along with `enableDataOptOut()` and `enableInputIconDisabling()` bascially all do the same. I would recommend re-factoring this to be DRY.

src/popup.html
```diff
@@ -215,6 +215,13 @@ <h2 class="fx-relay-panel-header-title i18n-content" data-i18n-message-id="popup
                 <span class="fx-relay-settings-toggle-label i18n-content" data-i18n-message-id="popupSettingsCollectData"></span>
                 <button title="" data-icon-visibility-option="disable-input-icon" class="data-collection fx-relay-settings-toggle"></button>
               </div>
+              <div class="fx-relay-settings-toggle-wrapper">
+                <!-- TODO: Localize string -->
+                <span class="fx-relay-settings-toggle-label xi18n-content" data-i18n-message-id="popupSettingsCollectData">
+                  Password manager compatibility mode
```
maxxcrawford, line 221: blocking: Localize string

src/js/other-websites/icon.js
```diff
@@ -19,8 +19,15 @@ async function run() {
   if (!inputIconsAreEnabled) {
     return;
   }
+
+  const pwdMgrCompat = await isPwdMgrCompatEnabled();
+
+  if (pwdMgrCompat) {
+    // TODO: Set RELAY_ICON_WIDTH width, hover state calcs
```
maxxcrawford, line 26: This is where the "rest of the work" is for this PR. This PR mostly gets a pref toggled to do the offest, rather than doing it.

maxxcrawford (commented):
self-review for future travelers

PR #530, opened 2023-10-09
Update add-on to non-persistent pages

maxxcrawford:
This requires adding the `"persistent": false` to the manifest.json and removing and global variables from the background script. Luckily, we only had one: RELAY_SITE_ORIGIN. This commit moves it to local storage (and is set on start-up) and adjusted any other time its referenced to rely on localStorage for it.

### DIFF
--- a/src/js/background/background.js
+++ b/src/js/background/background.js
@@ -1,8 +1,13 @@
-const RELAY_SITE_ORIGIN = "http://127.0.0.1:8000";
+function startupInit() {
+  const RELAY_SITE_ORIGIN = "http://127.0.0.1:8000";  
+  browser.storage.local.set({ RELAY_SITE_ORIGIN });
+  browser.storage.local.set({ maxNumAliases: 5 });
+  browser.storage.local.set({ relaySiteOrigin: RELAY_SITE_ORIGIN });
+  browser.storage.local.set({ relayApiSource: `${RELAY_SITE_ORIGIN}/api/v1` });
+}
 
-browser.storage.local.set({ maxNumAliases: 5 });
-browser.storage.local.set({ relaySiteOrigin: RELAY_SITE_ORIGIN });
-browser.storage.local.set({ relayApiSource: `${RELAY_SITE_ORIGIN}/api/v1` });
+browser.runtime.onStartup.addListener(startupInit);
+
 
 browser.runtime.onInstalled.addListener(async (details) => {
   const { firstRunShown } = await browser.storage.local.get("firstRunShown");
@@ -249,6 +254,7 @@ async function sendMetricsEvent(eventData) {
 
   const ga_uuid = await getOrMakeGAUUID();
   const eventDataWithGAUUID = Object.assign({ ga_uuid }, eventData);
+  const { RELAY_SITE_ORIGIN } = await browser.storage.local.get("RELAY_SITE_ORIGIN");
   const sendMetricsEventUrl = `${RELAY_SITE_ORIGIN}/metrics-event`;
   fetch(sendMetricsEventUrl, {
     method: "POST",
@@ -294,6 +300,8 @@ async function refreshAccountPages() {
 async function makeDomainAddress(address, block_list_emails, description = null) {
   const apiToken = await browser.storage.local.get("apiToken");
 
+  const { RELAY_SITE_ORIGIN } = await browser.storage.local.get("RELAY_SITE_ORIGIN");
+  
   if (!apiToken.apiToken) {
     browser.tabs.create({
       url: RELAY_SITE_ORIGIN,
@@ -360,8 +368,10 @@ async function makeDomainAddress(address, block_list_emails, description = null)
 // eslint-disable-next-line no-redeclare
 async function makeRelayAddress(description = null) {
   const apiToken = await browser.storage.local.get("apiToken");
-
+  
   if (!apiToken.apiToken) {
+    const { RELAY_SITE_ORIGIN } = await browser.storage.local.get("RELAY_SITE_ORIGIN");
+    
     browser.tabs.create({
       url: RELAY_SITE_ORIGIN,
     });
@@ -481,6 +491,8 @@ async function displayBrowserActionBadge() {
 browser.runtime.onMessage.addListener(async (m, sender, _sendResponse) => {
   let response;
 
+  const { RELAY_SITE_ORIGIN } = await browser.storage.local.get("RELAY_SITE_ORIGIN");
+
   switch (m.method) {
     case "displayBrowserActionBadge":
       await displayBrowserActionBadge();
@@ -540,8 +552,8 @@ browser.runtime.onMessage.addListener(async (m, sender, _sendResponse) => {
   return response;
 });
 
-
 (async () => {
+  startupInit();
   await displayBrowserActionBadge();
   await storeRuntimeData();
 })();
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -26,7 +26,8 @@
         "js/shared/utils.js",
         "js/background/background.js",
         "js/background/context-menu.js"
-      ]
+      ],
+      "persistent": false
   },
 
   "browser_action": {


src/js/background/background.js
```diff
@@ -1,8 +1,13 @@
-const RELAY_SITE_ORIGIN = "http://127.0.0.1:8000";
+function startupInit() {
+  const RELAY_SITE_ORIGIN = "http://127.0.0.1:8000";
```
pdehaan, line 2:
```suggestion
  const RELAY_SITE_ORIGIN = "http://127.0.0.1:8000";
```

src/js/background/background.js
```diff
@@ -1,8 +1,13 @@
-const RELAY_SITE_ORIGIN = "http://127.0.0.1:8000";
+function startupInit() {
+  const RELAY_SITE_ORIGIN = "http://127.0.0.1:8000";  
+  browser.storage.local.set({ RELAY_SITE_ORIGIN });
+  browser.storage.local.set({ maxNumAliases: 5 });
+  browser.storage.local.set({ relaySiteOrigin: RELAY_SITE_ORIGIN });
+  browser.storage.local.set({ relayApiSource: `${RELAY_SITE_ORIGIN}/api/v1` });
```
pdehaan, line 6: Ignorant Q: But do these have to be set separately, or can we set one object w/ multiple keys?
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/set implies we can almost condense these four `set()` calls into a single call w/ object w/ 4 keys. 🤷

src/js/background/background.js
```diff
@@ -1,8 +1,13 @@
-const RELAY_SITE_ORIGIN = "http://127.0.0.1:8000";
+function startupInit() {
+  const RELAY_SITE_ORIGIN = "http://127.0.0.1:8000";  
+  browser.storage.local.set({ RELAY_SITE_ORIGIN });
+  browser.storage.local.set({ maxNumAliases: 5 });
+  browser.storage.local.set({ relaySiteOrigin: RELAY_SITE_ORIGIN });
```
pdehaan, line 5: Why set the `RELAY_SITE_ORIGIN` twice w/ different aliases?

maxxcrawford: This is a good call-out. No good reason as to "why" but it's something we can/should fix. 

Filed #537

src/js/background/background.js
maxxcrawford: We can definitely consolidate! Filed #538

maxxcrawford:
@pdehaan Thanks for the previous review. I filed two issues (#537, #538) to capture/TODO them.

src/js/background/background.js
```diff
@@ -478,70 +488,79 @@ async function displayBrowserActionBadge() {
...
-  }
-  return response;
-});
+browser.runtime.onMessage.addListener((m, sender, sendResponse) => {
+  (async () => {
+    let response = null;
+
+    const { RELAY_SITE_ORIGIN } = await browser.storage.local.get("RELAY_SITE_ORIGIN");
+
+    switch (m.method) {
```
maxxcrawford, line 497: This approach is outlined in the [MDN documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage).

src/js/background/context-menu.js
```diff
@@ -519,7 +519,12 @@ const relayContextMenus = {
   },
 };
 
-// Events
+(async () => {
+  // Only show is platform support context menus
```
maxxcrawford, line 523: Android logic check

src/js/other-websites/inpage_menu.js
```diff
@@ -146,7 +146,7 @@ async function fillTargetWithRelayAddress(generateClickEvt) {
 
   if (serverStoragePref) {
     // Update server info with site usage
-    await browser.runtime.sendMessage({
+    browser.runtime.sendMessage({
```
maxxcrawford, line 149: Performance fix: dropped the `await` here as it shouldn't block the user from injecting the mask into the field.

maxxcrawford (commented):
Self review.

rafeerahman:
Ran the add-on from this branch in android.

(Blocking)

Running into one issue, the in-page menu is stuck on loading. 

image

Can confirm that running the add-on panel normally works as intended, as well as generating masks.

image

src/js/other-websites/inpage_menu.js
lloan: Nice!

config-chrome.sh
```diff
@@ -1,10 +1,11 @@
 #!/bin/bash
 
 # script to run the sed commands in both OSX and linux (e.g., GitHub Action)
-# replace "menus" permission (Firefox) with Chrome-compability permission ("contextMenus")
+# remove "contextMenus" permission (Firefox-only) from manifest.json
```
rafeerahman, line 4: This code is removing "menus" from manifest.json when I do `npm run web-ext-run:chrome` and whenever the scripts are run in github actions. Removing "menus" would cause the firefox context menu to not work, unless I manually add it back. 

Is there a reason for this and is this intended? Should we have a config.sh for firefox seperately that removes contextMenus and re-adds menus when running the firefox extension?

In the builds from github actions and the one tested in the original bug report https://github.com/mozilla/fx-private-relay-add-on/actions/runs/6844016974 ([MPP-3581](https://mozilla-hub.atlassian.net/browse/MPP-3581)), if you download the zip file you can see that only "contextMenus" is in the manifest.json. This is might be the reason why QA ran into the bug, since Firefox needs "menus". 

Is there a reason why we can't leave both menus and contextMenus in the manifest.json file?

Also confirming that the context menu works, but only when "menus" is in the manifest.json file.

src/js/background/background.js
```diff
@@ -478,70 +488,79 @@ async function displayBrowserActionBadge() {
...
-    case "updateAddOnAuthStatus":
-      await updateAddOnAuthStatus(m.status);
-      break;
-    case "updateInputIconPref":
-      browser.storage.local.set({ showInputIcons: m.iconPref });
-      break;
-  }
-  return response;
-});
+browser.runtime.onMessage.addListener((m, sender, sendResponse) => {
```
rafeerahman, line 491: In MPP-3490, when it says "Ensure listeners are registered synchronously at the top-level", what exactly does this mean? Do all listeners need to be registered synchronously? If so there are many others that don't follow this format.

maxxcrawford: If you look at what we [currently have](https://github.com/mozilla/fx-private-relay-add-on/blob/main/src/js/background/background.js#L489) on background.js, you'll see it's an async message. 

This same change is required for migrating to Manifest v3, so they're somewhat related.

maxxcrawford: I see the three you mean: 
- `browser.contextMenus.onClicked.addListener`(`background.js`)
- `browser.menus.onShown.addListener` (`context-menu.js`)
- `browser.runtime.onInstalled.addListener` (`context-menu.js`)

Specifically,  the two items in context-menu don't apply as that code/logic is not ran on Firefox for Android, but the start-up def. is, and may be impacting the mobile first-run. 

Solid call out. I'll adjust the `browser.runtime.onInstalled.addListener` function accordingly.

config-chrome.sh
maxxcrawford: Ahh! I should've flagged this change better in the self-review / code comments. 

The permission, [`menus`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/menus) does not exist for Chrome. Previously (before Oct 2023), we could ship permissions that Chrome did not recognize, check if they exist, and use them as available. However, Chrome rejected our submission and needed a clean (read: no declared `menus` permission) manifest file to approve our release. 

When we did this change, we re-submitted and Chrome accepted, etc. (and special note that we did NOT submit a Firefox release with this new change) 

QA flagged the following issue on the Firefox side: https://mozilla-hub.atlassian.net/browse/MPP-3581 where if you didn't ship BOTH permissions to Firefox, _it_ would break. 

That put us in a pickle. Either refactored a bunch of the code everywhere `.menus` and `.contextMenus` is mentioned OR ship both permissions to Firefox and only `contextMenus` to Chrome. 

This change to both the `config-chrome.sh` file and the `manifest.json` file (where both permissions are now included) makes the add-on work across platform.

src/js/background/background.js
rafeerahman: Oh in MPP-3490, by "top-level" did it mean the ones in background.js? I was also wondering about other listeners in different files.

src/manifest.json
```diff
@@ -93,6 +93,8 @@
     }
   ],
 
+  "content_security_policy": "object-src data: 'unsafe-eval'",
```
maxxcrawford, line 96: @Vinnl @groovecoder I added this to fix [MPP-3607](https://mozilla-hub.atlassian.net/browse/MPP-3607) to allow this bit of the icon injection script: https://github.com/mozilla/fx-private-relay-add-on/blob/main/src/js/other-websites/icon.js#L69-L70 

Is this unsafe? What are our options?

src/js/background/background.js
maxxcrawford: Yeah - Any `browser.foo` level listeners.

src/manifest.json
lloan: My two cents: If the SVG image source (iconUrl in your code) were to be constructed dynamically based on user input or untrusted data, it could be vulnerable to data injection attacks like Cross-Site Scripting (XSS). An attacker could potentially inject malicious code into the SVG data, leading to security vulnerabilities. That is not the case here. 

Maybe @groovecoder or @Vinnl can come up with another reason (or set of them).

Vinnl: For the actual risks I'd defer to security experts (e.g. in `#security`), but as restrictive as possible should be safest, and as far as I can see, it can be quite a bit tighter than this? Specifically, I don't see why the referenced line would need `'unsafe-eval'`, and I'd expect [`img-src`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/img-src) rather than `object-src`?

And since we already know what image we'll be injecting, for bonus points I think you should even be able to [specify the image's hash](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy#other_values), rather than just allowing all `data:` URLs. (But when the CSP is restricted to just images, I think the main danger is that malicious scripts could inject misleading images if our extension is installed, which seems fairly manageable.)

maxxcrawford: Welp, it wasn't loading on that page before the CSP policy, and it was loading after it. I'll play with some other policies (including the hash method too) to see how tightly I can get it to still work.

maxxcrawford: Update: I removed it and it's now _working_. 🤔

src/js/other-websites/icon.js
```diff
@@ -62,12 +62,11 @@ function wireUpInputs(emailInputs) {
...
-    );
+    // Rather than inject data: object, use CSP-safe SVG path
+    const iconURL = browser.runtime.getURL("/icons/icon-relay-inpage.svg");
```
lloan, line 66: Awesome - much better!

src/manifest.json
maxxcrawford: Ugh. Okay - So I refactored the add-on to use [`browser.runtime.getURL(path/to/icon.svg)`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getURL) to get CSP-save/declared path and inject that image. 

However, I don't think this will work. See more in Jira: https://mozilla-hub.atlassian.net/browse/MPP-3607

src/js/other-websites/inpage_menu.js
```diff
@@ -223,7 +223,7 @@ async function populateFreeMaskList(maskList, masks) {
     .querySelector(".fx-relay-menu-masks-lists")
     ?.classList.add("is-visible");
 
-  await browser.runtime.sendMessage({
+  browser.runtime.sendMessage({
```
rafeerahman, line 226: Nice, since the top-level listener is synchronous this seems to solve the issue

src/js/popup/popup.js
```diff
@@ -161,6 +161,16 @@
       }
     },
     init: async () => {
+
+      const { isAndroid } = await browser.storage.local.get(
+        "isAndroid"
+      );
+
+      // Add CSS class for custom Firefox for Android styles to the panel
+      if (isAndroid) {
```
rafeerahman, line 170: Confirming the height fills the screen on android.

src/js/background/background.js
```diff
@@ -1,28 +1,38 @@
...
+}
 
-browser.storage.local.set({ maxNumAliases: 5 });
-browser.storage.local.set({ relaySiteOrigin: RELAY_SITE_ORIGIN });
-browser.storage.local.set({ relayApiSource: `${RELAY_SITE_ORIGIN}/api/v1` });
+browser.runtime.onStartup.addListener(startupInit);
 
-browser.runtime.onInstalled.addListener(async (details) => {
-  const { firstRunShown } = await browser.storage.local.get("firstRunShown");
+browser.runtime.onInstalled.addListener((details) => {
```
rafeerahman, line 11: Great

src/manifest.json
groovecoder: Agree with what @lloan and @Vinnl said that it's best to find the tightest possible CSP. I also couldn't reproduce the bug that says the icon is NOT showing up on Mozilla's Auth0 SSO page in Mac Firefox, so it looks like this is a Chrome-only issue. The `getURL` technique seems right, but good catch that it would expose the extension's ID to the page, which is a potential privacy leak.

The only fix I can think of would be to modify the `Content-Security-Policy` HTTP response header of the web page itself to allow inline `background-image`s - i.e., NOT this `content_security_policy` in the add-on's manifest. And we shouldn't modify the CSP of all web pages to allow unsafe inline images just to make our add-on icon show up!

So yeah, I think we can remove this line, and we have to just leave the image broken on some pages.

```suggestion
```

groovecoder (changes requested):

src/js/popup/popup.js
maxxcrawford: It does not. I have #544 still out. We'll need to add extra CSS keyed off of the `is-android` classname. Width issue is fixed, though.

rafeerahman: Oops, I meant to say the width not the height 😵‍💫

